PR #1, opened 2020-06-26
fix: make code in build directory runnable

blakebyrnes:
This PR makes a few changes:
- build directory is runnable:
 - copies in the mitm/dist and replay/dist so it can be run after
 - you have to manually run yarn inside build once
- Removed shared/logger - moved to commons/Logger
- Create emulator-plugins-shared project to access the shared code instead of navigating backwards


blakebyrnes:
Additional modifications:
- tests run build directory
- replay now builds typescript into build directory and works with those files
- copyfiles is used to copy non-typescript assets into the build project

core/lib/DomEnv.ts
```diff
@@ -154,10 +154,10 @@ const typesonRegistryScript = fs.readFileSync(
 );
 
 const domStorageScript = fs.readFileSync(
-  require.resolve(`@secret-agent/injected-scripts/dist/scripts/domStorage.js`),
+  require.resolve(`@secret-agent/injected-scripts/scripts/domStorage.js`),
```
calebjclark, line 157: Does this show errors in WebStorm since it doesn't know about js build files from within Typescript files?

injected-scripts/tsconfig.json
```diff
@@ -5,10 +5,12 @@
...
     "rootDir": ".",
-    "outDir": "./dist",
+    "outDir": "../build/injected-scripts",
     "composite": true,
+    "resolveJsonModule": true,
+    "moduleResolution": "node",
     "removeComments": true
   },
-  "include": ["interfaces/*.ts", "scripts/*.ts"],
+  "include": ["./interfaces/*.ts", "./scripts/*.ts", "./package.json"],
```
calebjclark, line 14: That's cool. I didn't know non-TS files could be included from tsconfig.json!

replay/tsconfig.json
```diff
@@ -1,9 +1,10 @@
 {
   "compilerOptions": {
-    "outDir": "dist",
+    "outDir": "../build/replay",
     "module": "esnext",
     "target": "es2018",
     "moduleResolution": "node",
+    "composite": true,
```
calebjclark, line 7: Why is this needed?

calebjclark: I read up on `composite` in the TS documentation, but I"m not quite sure why we need it here.

blakebyrnes: Using the project references requires it

core/lib/DomEnv.ts
blakebyrnes: It doesn't right now

PR #2, opened 2020-06-26
feat: Prepare for publishing

blakebyrnes:
This PR adds a typescript build into a "build-dist" directory in preparation to publish out of:
- tsconfig.dist.json
- a postprocessor to update common package.json attributes and licenses
- an ability to customize output package.json settings using a package.dist.json in each package.

This pr also moves the postinstall script for mitm and emulators into the respective projects.


core-interfaces/package.json
```diff
@@ -1,16 +1,9 @@
 {
   "name": "@secret-agent/core-interfaces",
-  "version": "0.2.8",
+  "version": "1.0.0-alpha.0",
   "description": "Core interfaces used by SecretAgent",
-  "author": "Data Liberation Foundation",
-  "contributors": [
-    "Caleb Clark",
-    "Blake Byrnes"
-  ],
```
calebjclark, line 9: Why cut out author and contributors?

blakebyrnes: I changed it so we merge these all in during the build process so that we don't maintain in all these sub-packages.

PR #3, opened 2020-06-29
Website deployment

calebjclark:
- Connected documentation search
- Fixed Assets page for downloading logos/etc
- Website now builds to build-dist for deployment
- `yarn deploy` deploys to Github Pages (secretagent.dev)


website/src/components/SearchForm.vue
```diff
@@ -33,9 +33,9 @@
 
         import('docsearch.js').then(({ default: docsearch }) => {
           docsearch({
-            indexName: 'gridsome',
+            indexName: 'secretagent',
             inputSelector: `#${this.id}-input`,
-            apiKey: 'a7400a3a94b256c5283cb05efb860fc1',
+            apiKey: 'b975ccd9d9a93db63dd14f8d0baee5eb',
```
blakebyrnes, line 38: Are we supposed to check these in?

website/src/pages/Why.vue
```diff
@@ -33,7 +33,7 @@ BasicLayout.Why
         <a href="https://gs.statcounter.com/browser-version-market-share/desktop/worldwide/">top 3</a> most popular browsers are ready to plug-in.
 
 
-      h2 Writing Scraper Scripts is Still too Complicated
+      h2 Writing Scraper Scripts Is too Complicated
```
blakebyrnes, line 36: Why capitalized? Shouldn't that be lower case?

website/src/components/SearchForm.vue
calebjclark: Yes, it's the public api key.

website/src/pages/Why.vue
calebjclark: Verbs should always be capitalized. And according to https://titlecaseconverter.com/, the word "too" should also be capitalized... just pushed an update.

PR #7, opened 2020-06-30
fix: .gitignore was ignoring files that were needed for website

calebjclark:
Now that tsc is building js files into build/, we can remove some of the .gitignore lines. Doing so highlighted the fact that two needed website js files were being excluded.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,32 +6,19 @@ lerna-debug.log
 
 coverage
 .DS_Store
-dist
-build/
-website/.cache
 tsconfig.tsbuildinfo
 
 .idea/
 
 .cache-test
+
+/build/
 /build-dist/
 
-*.js
-
-!jest-*.js
-!dist-prepare.js
-
 /mitm/socket/connect
-!mitm/install.js
 /mitm/socket/build/
-/.mitm-ca
-/**/.mitm-ca
 
 emulators/data/DOM-*.json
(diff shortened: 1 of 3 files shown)


.gitignore
```diff
@@ -6,32 +6,18 @@ lerna-debug.log
...
 
-!jest-*.js
-!dist-prepare.js
+/build/
+/build-dist/
 
 /mitm/socket/connect
-!mitm/install.js
 /mitm/socket/build/
-/.mitm-ca
```
blakebyrnes, line 27: shouldn't this one still be here?

.gitignore
```diff
@@ -6,32 +6,18 @@ lerna-debug.log
...
-/**/.mitm-ca
 
 emulators/data/DOM-*.json
-!/emulator-plugins/shared/injected-scripts/*.js
-!/emulator-plugins/shared/test/inspectHierarchy.js
-/emulator-plugins/shared/injected-scripts/index.js
-
 
-/.sessions
```
blakebyrnes, line 36: Shouldn't this also still be .gitignored?

.gitignore
calebjclark: Yes. I started the new .secret-agent data dir as part of this same PR, then split.

Will push a fix.

.gitignore
calebjclark: Yep.

PR #22, opened 2020-07-04
Various Fixes for Getting the Double Agent tests running

blakebyrnes:
This PR adds a number of changes for fixing double agent tests:
- Large request bodies timing out
- Performance of page recorder too slow to load page
- Misidentifying XHR requests
- Not tracking websocket messages correctly for unknown plugins
- Emulators need to be able to set a specific user agent for Double Agent tests
- Update docker deployment option
- Fix mitm cleanup of closed sockets
- Handle a few polyfill issues


blakebyrnes:
NOTE: the commits are a bit off in this PR because this branch got into a merge loop because of tags for alpha 1 and alpha 2 that I couldn't untangle.

PR #28, opened 2020-07-10
Show commands in playbar

blakebyrnes:
This PR adds the following features to replay:
- Functional play/pause/playbar offset
- Command Overlay shows the current command and result if applicable (NOTE: these are currently core commands, not what is sent to client)
- Translation of core/Commands for execJsPath and default other ones.
- Highlight elements on page as the replay plays


replay/backend/models/TabBackend.ts
```diff
@@ -1,16 +1,20 @@
...
 import ReplayApi from '~backend/ReplayApi';
+import IRectangle from '~shared/interfaces/IRectangle';
+import Rectangle = Electron.Rectangle;
+import Application from '~backend/Application';
+import ipcMain = Electron.ipcMain;
+import ICommandResult from '~shared/interfaces/ICommandResult';
+import { ITick } from '~shared/interfaces/ISaSession';
 
-const domReplayerScript = Path.join(app.getAppPath(), '..', 'injected-scripts/domReplayer.js');
+const domReplayerScript = require.resolve('../../injected-scripts/domReplayer');
```
calebjclark, line 17: Why the change? Just curious.

blakebyrnes: It didn't work with the new build directories when using dev mode

PR #29, opened 2020-07-13
Add interaction events to replay

blakebyrnes:
This PR adds the following interactions to replay:
1) Mouse events
2) Focus events (they're shown as "highlighted nodes" so that we don't try to steal the user's mouse focus)
3) Scroll events
4) Fix multi-page flows
5) Menus for changing date/session page now work.
6) Auto-play session when launched from full-client

Log Capture:
1) Capture logs to SessionLogs Table
2) Detect unhandled errors and send to core
3) Add client to stack traces 
4) Detect no commands in 5 seconds and errors and tell user
5) UX Tweaks

NOTE: this PR does NOT address adjusting the viewport of the page to match the script.

One other thing worth understanding - the current playback is simply running a loop and going 0.1 out of 100 forward every 10 ms. This means it's possible in the current strategy to have certain events on the timeline that are too fine-grained to pick up - ie, if an event occurs < a ratio of 0.1/100 applied to the ms of the script after another one, the "window" to show it is too small to hit. I think as we make this better, we'll likely want to adjust how we spread out the ticks a bit.


commons/Logger.ts
```diff
@@ -1,54 +1,65 @@
...
-    if (this.level === 'stats' || this.level === 'info') {
-      console.log(
-        `${new Date().toISOString()} INFO ${action}`,
-        ...[parentMessageId, data].filter(x => x !== undefined && x !== null),
-      );
-      return (logId += 1);
-    }
+  public info(sessionId: string, action: string, data?: any, parentLogId?: number) {
+    return this.log('info', sessionId, action, data, parentLogId);
   }
```
calebjclark, line 18: It seems these args types open the door for mistakes. For example, `log.stats(action, data)` is valid so long as data is a string. I believe this log item would then drop through the cracks since `action` matches no valid `sessionId`.

Perhaps this isn't much of a concern.

Regardless, I'm not sure of a legit way to fix this other than switching sessionId to be a `number`, which would be a broad undertaking.

core/lib/ChromeCore.ts
```diff
@@ -93,9 +96,9 @@ export default class ChromeCore {
 
         const pages = await puppBrowser.pages();
         await Promise.all(pages.map(async x => x.close())).catch(error => {
-          log.warn('error closing initial puppeteer browser page', error);
+          log.warn('Error closing initial chrome browser page', error);
```
calebjclark, line 99: Isn't this log.warn missing a sessionId arg?

And how was it not caught during typescript build? Or is `error` being treated as a string type?

commons/Logger.ts
blakebyrnes: It's a good point. I just tried explicitly making data an object. This helps a small amount, and found a few log problems that were in the current pr. I don't think it makes it impossible to send in a bad log though. I think I would need to move more towards sessionId being an optional param off of data for that to be the case. Or maybe if you send data, it has to include a sessionId if you send it (so you'd have to explicitly say null)

core/lib/ChromeCore.ts
blakebyrnes: Yep, it's a bug. Error is type "any" in a catch, for some reason.

commons/Logger.ts
blakebyrnes: Based on this, I decided to change the logging signature to have action, then a data object that, if passed, requires you to say what sessionId the logs are associated with.

PR #45, opened 2020-08-14
Replay feed

blakebyrnes:
This PR changes the way Replay runs when developing. Changes:
1) Replay API is moved into session-state, and now uses an http2 push api
2) SessionState tables have a "subscribe" function that will publish all data writes to the subscriber
3) Session Ticks in Replay-App are simplified to calculate previous mouse/focus/paint in an "upfront loop"
4) Replay resources are synched to Replay App and "waited for" by the frontend - this fixes all the resource not-found errors and approximates more closely the true session - aka, resources are still pending.


client/lib/Browser.ts
```diff
@@ -200,7 +200,12 @@ export async function createBrowser(
...
-  scriptInstance.launchReplay(sessionName, sessionsDataLocation);
+  let showReplay = envShowReplay;
+  if (options.showReplay !== undefined) showReplay = options.showReplay;
+  if (showReplay) {
+    scriptInstance.launchReplay(sessionName, coreClientSession);
+  }
   return new Browser(coreClientSession, sessionName);
 }
+
+const envShowReplay = Boolean(JSON.parse(process.env.SA_SHOW_REPLAY ?? 'true'));
```
calebjclark, line 211: Why create this `envShowReplay` const? It's only used once, on line 203. Why not just move this value to 203?

Regardless, it seems a bit odd to declare a const at the bottom.

blakebyrnes: Just because it has to parse the string to translate to a Boolean, so it seemed like it made sense to read it in once. You find that odd?

core-interfaces/AttachedState.ts
```diff
@@ -0,0 +1,2 @@
+const getAttachedStateFnName = '__getSecretAgentNodeState__';
+export default getAttachedStateFnName;
```
calebjclark, line 2: Why give this filename a different name than the default export?

client/lib/Browser.ts
calebjclark: In some circumstances I wouldn't, but in this case it's only being used in the createBrowser method, which isn't a commonly used method (i.e., like a `log` method that might be called thousands of times in the course of a single script). It complicates the code for a trivial (practically non-existent) efficiency boost.

blakebyrnes:
It should probably be changed. I mostly was just moving it out of core so I could reference it without needing to pull in core

core/index.ts
```diff
@@ -21,10 +21,14 @@ import UserProfile from './lib/UserProfile';
...
 import { IRequestInit } from 'awaited-dom/base/interfaces/official';
 import IAttachedState from '@secret-agent/injected-scripts/interfaces/IAttachedStateCopy';
-import Signals = NodeJS.Signals;
 import Log from '../commons/Logger';
+import { createReplayServer } from '@secret-agent/session-state/api';
+import ISessionReplayServer from '../session-state/interfaces/ISessionReplayServer';
+import Signals = NodeJS.Signals;
 
 const { log } = Log(module);
+const shouldStartReplayServer = Boolean(JSON.parse(process.env.SA_SHOW_REPLAY ?? 'true'));
```
calebjclark, line 30: In this case, we're parsing the env var at the top of the file, which seems to make a lot more sense, even if it's only used once.

PR #47, opened 2020-08-18
Windows tests

blakebyrnes:
This PR adds automated tests for windows, and changes the way we cleanup after tests to try to handle more scenarios where things blow up. As a part of this PR, the servers that we use that are only exposed internally are now seeking out their own open ports by default


core/lib/GlobalPool.ts
```diff
@@ -37,9 +32,6 @@ export default class GlobalPool {
   }[] = [];
 
   public static async start() {
-    if (this.isStarted) {
-      throw new Error('GlobalPool is already started');
-    }
```
calebjclark, line 42: Why get rid of `isStarted`?

blakebyrnes: We weren't using it anymore. GlobalPool no longer had a single startup requirement in the code. I don't know when we switched to booting based on if a chromeCore instance existed or not, but that's what the code does. I think this is legacy from Ulixee. 

I was tracking down an issue in Windows where it was still shutting down during the next startup, and cleared this out when I saw it didn't actually do anything anymore. We never set isStarted to true.

core/lib/GlobalPool.ts
```diff
@@ -90,17 +82,12 @@ export default class GlobalPool {
...
     }
-    if (this.chromeCore) {
-      await this.chromeCore.close();
-    }
-    if (this.mitmServer) {
-      await this.mitmServer.close();
-    }
-    SessionsDb.shutdown();
-    this.waitingForAvailability = [];
+    this.waitingForAvailability.length = 0;
```
calebjclark, line 85: Why this change to `length = 0`?

blakebyrnes: It's probably actually unnecessary. I was just keeping the same instance instead of potentially creating a new one in between "awaits" for closing.

core/lib/GlobalPool.ts
calebjclark: Ok, yeah... crazy that it was doing nothing.

PR #49, opened 2020-08-19
feat(mitm): support push streams

blakebyrnes:
This PR refactors the Mitm project to have an http2 server in the middle for https so that we can support push streams.


core/lib/GlobalPool.ts
```diff
@@ -98,7 +98,9 @@ export default class GlobalPool {
   ): Promise<Session> {
     this._activeSessionCount += 1;
     try {
-      return await this.chromeCore.createSession(options);
+      const session = await this.chromeCore.createSession(options);
+      if (!session) this._activeSessionCount -= 1;
+      return session;
```
calebjclark, line 103: createSession can return no session and yet not throw an error?

blakebyrnes: I wasn't sure what to do here... this will only happen during shutdown on a pending creation.

calebjclark: I'm looking at the createSession in ChromeCore... it seems our current setup will have weird side affects... not sure what to do. It seems weird that createSession would return undefined... what happens to the dev's code that is expecting a session back? Should an error be thrown? 

This PR probably isn't the one to fix it.

PR #50, opened 2020-08-21
fix(replay): fix rendering doctype + svg

blakebyrnes:
This PR fixes Replay so it can display namespaced attributes and elements (ie, svg).

It also moves testing to github actions and resolves a few issues on Ubuntu and Windows.


core/index.ts
```diff
@@ -249,39 +253,55 @@ export default class Core implements ICore {
...
+      }
+      const window = session.window;
+      this.byWindowId[window.id] = new Core(session);
+      return {
+        sessionId: session.id,
+        sessionsDataLocation: session.baseDir,
+        windowId: window.id,
+        replayApiServer: this.replayServer?.url,
+      };
+    });
```
calebjclark, line 270: Why are we putting this in a Queue?

blakebyrnes: There are several async actions in each of the shutdown, startup, createsession, etc functions in here. We now additionally have a delayed "shutdown" when you close a session and you didn't manually start. When you shutdown and start frequently, it's created too many problems with a shutdown happening "mid-session creation" and other similar problems

blakebyrnes: A queue is a strange name for this, but just checking a variable at the beginning of each function isn't working. Any ideas for a better name of the queue or a different pattern?

calebjclark: Yeah, the naming isn’t very clear what’s its doing. I don’t have a better idea right now.

PR #52, opened 2020-08-24
Emulator browsers

blakebyrnes:
- Emulators now define an "engine" they run on
- We download the "revision #" for each chromium instance to a shared folder on the machine that mimics what Playwright is doing
- Global Pool now manages multiple chrome instances (one for each revision)
- Remove Chrome 78,79 emulators
- Add Chrome 83 emulator + support for new TLS signature


humanoids/index.ts
```diff
@@ -14,14 +14,13 @@ export default class Humanoids {
...
+  public static create(key: string) {
     return new this.pluginsByKey[key]();
   }
 
-  public static getRandom(): HumanoidPlugin {
-    const plugins = Object.values(this.pluginsByKey);
-    const plugin = Utils.pickRandom(plugins);
-    return new plugin();
+  public static getRandomId() {
+    const plugins = Object.keys(this.pluginsByKey);
```
calebjclark, line 22: `plugins` should be renamed to `pluginKeys`... or are we calling these `ids` now?

blakebyrnes: Ok, will fix.

PR #55, opened 2020-08-29
Replay loading on second page

blakebyrnes:
This PR modifies replay to fix an issue reloading the menus, and an issue where second pages are not displaying correctly sometimes. As part of this, I took a baby step in the direction of moving off mobx.

NOTE: at some point, going to need to overhaul how the replay is happening in browser. Basically, anytime you navigate, it's needing to "load" the second page before it can continue, but the state disappears between page loads, so we sometimes lose paint events.


replay-app/frontend/src/models/OverlayStore.ts
```diff
@@ -1,9 +1,8 @@
 import { ipcRenderer, remote } from 'electron';
-import { observable, computed } from 'mobx';
+import { computed, observable } from 'mobx';
```
calebjclark, line 2: 👀

blakebyrnes: Ha. Part of the eslint stuff

blakebyrnes: Sorts imports

PR #67, opened 2020-10-06
Http2 Bug Fixes; Don't shutdown on UnhandledRejection

blakebyrnes:
This PR fixes a few bugs:

Core/Client:
- Don't disconnect and shutdown on "unhandledPromiseRejection" since this doesn't exit the process
- Cancel pending CommandQueue items when we disconnect

Mitm:
- Http2 push headers were improperly handling request/response headers
- Properly handle "lists" of headers (ie, more than 1 entry)
- Don't manipulate http2 headers to look like http1 headers - we didn't have the http2 server flag set in time for this call

Replay:
- Hide replay elements using a shadow dom
- Reset navigation when going back and then selecting a new "replay" to view.

### DIFF
--- a/emulator-plugins/shared/injected-scripts/index.ts
+++ b/emulator-plugins/shared/injected-scripts/index.ts
@@ -18,18 +18,7 @@ export default class InjectedScripts {
   const sourceUrl = '${injectedSourceUrl}';
   ${utilsScript}
   
-  function runNewDocumentScript(){ 
-    ${this.scripts.join('\n\n')}      
-  }
-  // if main frame and HTML element not loaded yet, give it a sec
-  if (window.self === window.top && !document.documentElement) {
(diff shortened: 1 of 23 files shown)


emulator-plugins/shared/injected-scripts/index.ts
```diff
@@ -18,18 +18,7 @@ export default class InjectedScripts {
...
-  }
-  // if main frame and HTML element not loaded yet, give it a sec
-  if (window.self === window.top && !document.documentElement) {
-    new MutationObserver((list, observer) => {
-      observer.disconnect();
-      runNewDocumentScript();
-    }).observe(document, {childList: true, subtree: true});
-  } else {
-    runNewDocumentScript();
-  }
```
calebjclark, line 32: Why remove this?

PR #71, opened 2020-10-09
fix(core): wait for location change on new tab

blakebyrnes:
This PR changes a few things:
1) Connect redirects to navigations after the initial one (ie, new link, new tab)
2) waitForNewTab waits for an initial resource to be loaded - this is so that location "Change" events will not trigger for the first page loaded in a new tab.

### DIFF
--- a/core/index.ts
+++ b/core/index.ts
@@ -22,6 +22,7 @@ import ISessionReplayServer from '@secret-agent/session-state/interfaces/ISessio
 import Queue from '@secret-agent/commons/Queue';
 import Chrome83 from '@secret-agent/emulate-chrome-83';
 import Emulators from '@secret-agent/emulators';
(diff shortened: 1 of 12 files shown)


core/index.ts
```diff
@@ -64,7 +65,7 @@ export default class Core implements ICore {
   }
 
   public async goto(url: string) {
-    return this.tab.runCommand<void>('goto', url);
+    return this.tab.runCommand<IResourceMeta>('goto', url);
```
calebjclark, line 68: Is there really value in returning the resource?

blakebyrnes: This code change is simply correcting the typing, not changing the return type.

I do believe there's value if you are loading something other than an html page - ie, you're loading a json file.

PR #72, opened 2020-10-09
Small bug fixes

blakebyrnes:
1) Location change was still triggering on a new tab if you clicked a new link to get there and it opened in the background
2) Mitm will blow if client push session throws an error
3) Mitm still duplicating cache header


core/test/location.test.ts
```diff
@@ -399,7 +399,6 @@ setTimeout(() => {
     expect(history[3].stateChanges.has('AllContentLoaded')).toBe(true);
 
     expect(locationStatusHistory).toMatchObject([
-      'change',
```
calebjclark, line 402: Why is this being dropped?

PR #83, opened 2020-10-31
Screen Sizes + Locale emulation

blakebyrnes:
NOTE: this PR is stacked on #82. Once that's merged, I'll rebase this.

This PR adds the ability to emulate:
- Locale: sets locale in navigator and accept-language headers. The user can configure this in openBrowser
- TimezoneId: sets the timezone that will be used for dates/times
- Screen sizes: sets the screen size, browser window size, and location
   - if no screen size is provided, a screen will be allocated based on resolution statistics pulled from statcounter.com. Position on the screen and size of browser window will be randomized to fill 70%+ of the available screen.
   - NOTE: the platform specific "availWidth"/Height are not yet accounting for header size and dock sizes. I'll log a follow up ticket to explore that in double agent.

Replay supports screen sizes:
- It will emulate the screen size provided and scale up/down the window to match the screen width
- NOTE: height does not grow to fill, so there's a blank spot in the bottom of the page right now. I think we'll want to add another view to replay (or grow the replay bar) to cover this area with some kind of inset shadow indicating a margin. Or possibly this area would show information about the session (like screen size, emulator used, etc).


emulators/data/resolution.json
```diff
@@ -0,0 +1,146 @@
+{
+  "month": "2020-10",
+  "lastModified": "2020-10-30T15:58:20.354Z",
+  "sizes": [
+    {
```
calebjclark, line 5: Where did all this data come from?

calebjclark: Never mind. Bad question.

blakebyrnes: Statcounter

PR #90, opened 2020-11-11
Browser Emulators Clarification

blakebyrnes:
This PR adds clarity to browser emulator apis and makes them explicit for dom changes as opposed to mostly being hidden before.

Also adds docs on how to extend or write a browser emulator


emulate-browsers/base/injected-scripts/_descriptorBuilder.ts
```diff
@@ -0,0 +1,186 @@
+const nativeErrorRegex = new RegExp(/^(\w+):\s/);
+
+const globalSymbols = {};
+for (const symbol of ReflectCached.ownKeys(Symbol)) {
+  if (typeof Symbol[symbol] === 'symbol') {
+    globalSymbols[`${String(Symbol[symbol])}`] = Symbol[symbol];
+  }
```
calebjclark, line 7: Why does this filename begin with an underline?

blakebyrnes: I made the helper classes start with underlines just to indicate they weren't themselves "dom overrides"

PR #93, opened 2020-11-19
fix: emulators were failing some double-agent tests

calebjclark:
- Quality score is removed from locale property of emulators
- accept-language header bypasses using default values from DA json if emulator's locale has been set
- MediaRecorder.isTypeSupported correctly accepts the type arg
- SecretAgent.configure now accepts all properties of Core.configure
- Updated website docs

### DIFF
--- a/client/interfaces/ISecretAgentClass.ts
+++ b/client/interfaces/ISecretAgentClass.ts
@@ -4,7 +4,7 @@ import IUserProfile from "@secret-agent/core-interfaces/IUserProfile";
 import ICreateSecretAgentOptions from "./ICreateSecretAgentOptions";
 import ISecretAgent from "./ISecretAgent";
 
-export interface ISecretAgentConfigureOptions extends Pick<ICoreConfigureOptions, 'browserEmulatorIds'> {
+export interface ISecretAgentConfigureOptions extends ICoreConfigureOptions  {
   defaultRenderingOptions: IRenderingOption[];
   defaultUserProfile: IUserProfile;
 }
--- a/emulate-browsers/base/injected-scripts/MediaRecorder.isTypeSupported.ts
+++ b/emulate-browsers/base/injected-scripts/MediaRecorder.isTypeSupported.ts
@@ -1,7 +1,7 @@
 const { supportedCodecs } = args;
 
 if ((window as any).MediaRecorder) {
-  proxyFunction((window as any).MediaRecorder, 'isTypeSupported', (func, thisArg, type) => {
+  proxyFunction((window as any).MediaRecorder, 'isTypeSupported', (func, thisArg, [type]) => {
     if (type === undefined) return ProxyOverride.callOriginal;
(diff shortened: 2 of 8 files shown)


emulate-browsers/chrome-80/index.ts
```diff
@@ -46,7 +46,7 @@ export default class Chrome80 {
 
   public readonly userAgent: IUserAgent;
   public readonly networkInterceptorDelegate: INetworkInterceptorDelegate;
-  public locale = 'en-US,en;0.9';
+  public locale = 'en-US,en';
```
blakebyrnes, line 49: Does this not break the request headers?

emulate-browsers/base/injected-scripts/MediaRecorder.isTypeSupported.ts
```diff
@@ -1,7 +1,7 @@
 const { supportedCodecs } = args;
 
 if ((window as any).MediaRecorder) {
-  proxyFunction((window as any).MediaRecorder, 'isTypeSupported', (func, thisArg, type) => {
+  proxyFunction((window as any).MediaRecorder, 'isTypeSupported', (func, thisArg, [type]) => {
```
blakebyrnes, line 4: Does this blow up if you call MediaRecorder.isTypeSupported() with no params?

client/interfaces/ISecretAgentClass.ts
```diff
@@ -4,7 +4,7 @@ import IUserProfile from "@secret-agent/core-interfaces/IUserProfile";
 import ICreateSecretAgentOptions from "./ICreateSecretAgentOptions";
 import ISecretAgent from "./ISecretAgent";
 
-export interface ISecretAgentConfigureOptions extends Pick<ICoreConfigureOptions, 'browserEmulatorIds'> {
+export interface ISecretAgentConfigureOptions extends ICoreConfigureOptions  {
```
blakebyrnes, line 7: Should we update the docs if we open these up?

emulate-browsers/base/injected-scripts/MediaRecorder.isTypeSupported.ts
calebjclark: No

emulate-browsers/chrome-80/index.ts
calebjclark: Not in DA.

client/interfaces/ISecretAgentClass.ts
calebjclark: Probably.

emulate-browsers/chrome-80/index.ts
blakebyrnes: So if you set it this way, is Chrome still adding the q=0.9 that is on network requests? Or is the Double Agent test not checking Accept-Language?

PR #96, opened 2020-11-25
Emulator data

calebjclark:
- Uses new json data structure from Shapeshifter repo (I'm still trying out the name 😎).
- Replaced UserAgents.ts with user-agent-options.ts from Shapeshifter.
- The big downside to this PR is it downgrades http headers to only include Document type.
- Screen size is no longer randomly selected. Instead it defaults to 1440x900.
- Default viewport now takes into account the OS top bar and dock/start menus.


core-interfaces/IBrowserEmulator.ts
```diff
@@ -1,10 +1,10 @@
 import INetworkInterceptorDelegate from './INetworkInterceptorDelegate';
 import IUserProfile from './IUserProfile';
-import IUserAgent from './IUserAgent';
 import INewDocumentInjectedScript from './INewDocumentInjectedScript';
 
 export default interface IBrowserEmulator {
-  readonly userAgent: IUserAgent;
+  readonly navigatorUserAgent: string;
```
blakebyrnes, line 6: This user agent is used for network requests too right?

core-interfaces/IBrowserEmulatorClass.ts
```diff
@@ -1,11 +1,10 @@
-import IUserAgent from './IUserAgent';
 import IBrowserEmulator from './IBrowserEmulator';
 
 export default interface IBrowserEmulatorClass {
   id: string;
   roundRobinPercent: number;
   engine: { browser: string; executablePath: string; revision: string };
-  new (userAgent?: IUserAgent): IBrowserEmulator;
+  new (): IBrowserEmulator;
```
blakebyrnes, line 7: Assuming you have a way to set this in double agent?

emulate-browsers/base/injected-scripts/polyfill.additions.ts
```diff
@@ -1,7 +1,7 @@
 for (const addition of args.additions || []) {
   try {
     if (addition.propertyName === 'getVideoPlaybackQuality') {
-      addition.property['_value()'] = function() {
+      addition.property['_$$value()'] = function() {
```
blakebyrnes, line 4: Double $ on purpose? I'm actually fine with it, just curious

emulate-browsers/base/lib/getLocalOperatingSystemMeta.ts
```diff
@@ -0,0 +1,112 @@
...
+  darwin: 'mac-os',
+  win32: 'windows',
+  linux: 'linux',
+  aix: 'linux',
+  freebsd: 'linux',
+  openbsd: 'linux',
+  sunos: 'linux',
+}
+
+const darwinToMacOsVersionMap = {
```
blakebyrnes, line 72: Any thoughts on how to make sure this stays up to date? Is anyone tracking this in a high usage, no deps lib?? :)

emulate-browsers/chrome-80/data/mac-os-10-14/dom-diffs-when-using-mac-os-10-14.json
```diff
@@ -0,0 +1,177 @@
...
+        ]
+      }
+    }
+  ],
+  "order": [],
+  "changes": [
+    {
+      "path": "window.webkitMediaStream.new().id",
+      "propertyName": "_$value",
+      "property": "49e9ade4-99c6-4b15-a35b-593b7ae1bf23"
```
blakebyrnes, line 159: Thought these got removed?

emulate-browsers/chrome-80/data/mac-os-10-14/window-chrome.json
```diff
@@ -0,0 +1,454 @@
...
+        "w"
+      ]
+    },
+    "_$type": "object",
+    "_$flags": [
+      "e",
+      "w"
+    ]
+  },
+  "prevProperty": "Atomics"
```
blakebyrnes, line 453: Still no runtime?

emulate-browsers/chrome-80/index.ts
```diff
@@ -5,33 +5,29 @@ import {
...
+const codecsData = new DataLoader(`${__dirname}/data`, 'codecs');
+const domDiffsData = new DomDiffLoader(`${__dirname}/data`);
 
 @BrowserEmulatorClassDecorator
 export default class Chrome80 {
   public static id = pkg.name;
-  public static roundRobinPercent = StatcounterBrowserUsage.getConsumerUsageForBrowser(
-    'Chrome 80.0',
-  );
+  public static roundRobinPercent = 0; // TODO; Import from json
```
blakebyrnes, line 30: Are you planning to import the consumer browser usage from double agent?

emulate-browsers/base/data-scripts/updateScreenSizesStats.ts
```diff
@@ -1,52 +0,0 @@
-import moment from 'moment';
```
blakebyrnes, line 1: I agree we don't want to auto-rotate these, but I think we might want to keep the ability to periodically import the top screen sizes. I don't know where exactly we want this to live if it's turned off by default though... ie, how does a user get random sizes?

blakebyrnes: Maybe we need a separate package for randomizing values - like browsers, screensizes, timezone+languages, etc. If we're not going do turn these on by default, it might be nice to have a "randomizer" package to help do so.

calebjclark: The more I've thought about this the more I believe that it's highly unlikely a developer will ever want to randomize screen sizes. It's because of MediaQueries. They affect whether an element is even in the DOM, which seems to make randomization dangerous.

Also, how often do you think the top screen size changes? It's years, right?

emulate-browsers/chrome-80/index.ts
calebjclark: Shit. Yes.

emulate-browsers/base/data-scripts/updateScreenSizesStats.ts
blakebyrnes: I still think you're thinking about this edge-case in. Yes, it is possible that you might hide an element with a media query, but I think if you are rotating through common "desktop" screen sizes, the range that will impact an element you want to scrape is wider than the randomization. Ie, you will mostly see no changes to elements of importance rotating through desktop screen sizes. I don't have data for that, just going off of my experience using media queries. You hide elements that are things like nav, not the core data.

Separately, screen size is part of fingerprinting. I don't see how we think we can send across the same screen size, position, and browser window size every time and think that won't be detected.

blakebyrnes: Stats, fwiw: https://gs.statcounter.com/screen-resolution-stats/desktop/worldwide/#monthly-201910-202010

calebjclark: Anything that stood out to you in the stats?

calebjclark: Screen size is a very minor piece of fingerprinting. Look at the stat percentages. The bucket that your screen size places you in is vast. The screen size on my 15" MacBook Pro is the same size as almost everyone else who has the 15" Macbook Pro.

calebjclark: And... I don't know what "mostly" means when you say, "you will mostly see no changes to elements of importance rotating through desktop screen sizes"... if you have a scraper script that hinges something off of a navigation tab and the rotating screens means that your script breaks 10% of time... that's going to be a huge hit to your scraping project, and it's going to be reaaaaaaaaly frustrating to debug... and you're going to blame the issue on us, which you should.

emulate-browsers/chrome-80/index.ts
calebjclark: Fixed.

emulate-browsers/chrome-80/data/mac-os-10-14/window-chrome.json
calebjclark: Not yet.

emulate-browsers/chrome-80/data/mac-os-10-14/dom-diffs-when-using-mac-os-10-14.json
calebjclark: There were, but the branch didn't have all the latest files.

Added.

emulate-browsers/base/data-scripts/updateScreenSizesStats.ts
blakebyrnes: I guess you must see this very differently than I do. I don't think your script will break 10% of the time if you chose to implement a screen rotation strategy when using the top 4 resolutions on that list. My uninformed guess would be 0% for 95% of scripts. I guess I just don't think the web looks like a place where media queries are impacting desktop resolutions that drastically.

If we include mobile and/or orientation differences, I agree with your sentiment.

Two things stood out to me about the stats:
1) The top screen size represents 20% of traffic. We made 50% of traffic our goal for browsers, so it seemed like we might want to do the same here
2) The top 4 have a bigger width jump than height, but unless there's different settings for >1366 width and >766 width, I think this is likely the same bucket of media query resolutions. Hard to imagine the use cases where designers are selectively changing things for this large setup.

emulate-browsers/chrome-80/data/config.json
```diff
@@ -0,0 +1,4 @@
+{
```
blakebyrnes, line 1: I like the data dir, but I wonder if the engine and maybe config should be at top level?

calebjclark: Maybe, but it's a simple copy and paste from Shapeshifter right now.

emulate-browsers/base/lib/getLocalOperatingSystemMeta.ts
calebjclark: I just pushed a commit that includes the wiki URL where I grabbed the data. I didn't see a dep lib, and I'm not certain of the benefits of trusting a 3rd party for something that changes maybe once a year.

All the version matching (when getting the local operating system and when getting the dom diffs) use `getClosestNumberMatch` in VersionUtils.ts. This method gets the closest version match without requiring an exact match. This means, developers can upgrade to a new OS version without our code needing to change.

emulate-browsers/chrome-80/data/config.json
blakebyrnes: Oh, you're driving the engine from shapeshifter now. Gotcha.. I had that coming first from the emulator before. I'm not sure about this one. I guess it's fine this way. Just feels a little buried what engine is underneath, if that matters.

emulate-browsers/base/injected-scripts/polyfill.additions.ts
calebjclark: Yes, was trying to separate the custom meta keys used in the DA export from the custom keys being added in SA.

emulate-browsers/base/lib/getLocalOperatingSystemMeta.ts
blakebyrnes: Ok. Is there some kind of warning or error if we can't match? If this is a year apart for changes, seems very easy to forget to update it.

emulate-browsers/base/injected-scripts/polyfill.additions.ts
calebjclark: It was more just for my sanity. I'm not against dropping it to a single $ if you think it best.

core-interfaces/IBrowserEmulatorClass.ts
calebjclark: Yes. Each emulator has a static getter called userAgentOptions, which I override in DA (well, actually Scraper Report / Benchmarker) to select the user-agent string I want.

core-interfaces/IBrowserEmulator.ts
calebjclark: Yes.

emulate-browsers/base/lib/getLocalOperatingSystemMeta.ts
calebjclark: No. I'm not sure where to put the warning. I'm against clogging up Stdout with these types of messages.

core-interfaces/IBrowserEmulator.ts
blakebyrnes: Does the name navigator reflect that you were seeing it in window.navigator? Or are you thinking of "navigator" as the browser?

emulate-browsers/chrome-80/data/config.json
calebjclark: I agree. One of the next items on my plate is to create all the new emulators... I think that's going to shift the emulator structure a bit... might be a good time to change it then... right now it's nice to just copy and past the dir and know all the data files were copied.

emulate-browsers/base/lib/getLocalOperatingSystemMeta.ts
blakebyrnes: If there's no warning, and no message to console, this seems very likely, imo, to be lost. Could it be part of the shapeshifter somehow?

emulate-browsers/base/data-scripts/updateScreenSizesStats.ts
calebjclark: Let's discuss further on the phone... I agree with your guess of "0% for 95% of scripts". It's the 5% that bothers me because the 5% turns into a massive frustration that's incredibly difficult to debug for those 5%.

core-interfaces/IBrowserEmulatorClass.ts
blakebyrnes: This is the interface for the Class, right? Shouldn't we put it on here if you need every emulator to have it on the class?

emulate-browsers/base/data-scripts/updateScreenSizesStats.ts
calebjclark: An idea... 

I could see setting different screen sizes, but picking a browser width/height that can be consistent within those screen sizes..

calebjclark: It seems as of a few days ago we dropped our near-term target of 50% for browsers.... now it's just the single most popular browser.

blakebyrnes:
FYI - windows tests are all failing because the OS isn't supported... ha. Maybe you made it too accurate??!?

emulate-browsers/base/data-scripts/updateScreenSizesStats.ts
blakebyrnes: Another idea.. perhaps this is a finding mission-impossible intends to find out the answer to - the variety of media queries used in the wild

core-interfaces/IBrowserEmulatorClass.ts
calebjclark: No because the only emulators that need it are the ones we're directly using in Scraper Report, which is our emulators. It shouldn't be required for "other developers" (if it ever comes to that... ha).

core-interfaces/IBrowserEmulator.ts
calebjclark: It reflects window.navigator. I can see an argument both for/against it.

emulate-browsers/base/lib/getLocalOperatingSystemMeta.ts
calebjclark: Maybe. Not quite sure how yet.

calebjclark: I'm not sure pulling it into Shapeshifter solves anything...

The data is still manually gathered... so I'm not sure how we would be "automatically" notified when the data is out of data. We might be able to solve this when we add "analytics" to SA.

core-interfaces/IBrowserEmulatorClass.ts
blakebyrnes: I have a feeling we will internally forget... the interface seems just as much for us as others, but I don't have anything else to add if you don't agree.

calebjclark: Ok. I added it.

blakebyrnes (approved):

PR #97, opened 2020-11-27
Bug fixes + Upstream Proxy Configuration

### LINKED ISSUE #92: set different proxies for different tabs?
is it possible to set different proxies for different tabs using secret agent?

blakebyrnes:
This PR has a few fixes, and a simplification for configuring upstream proxies.

1) Goto errors were not bubbling back to the client, so you would not see that anything broke. This is fixed in this PR so that navigation errors will throw back to the originating client goto call.
2) Replay "loads" of urls were occasionally stalling (an underlying electron no-op issue, I think). This PR works around those stalls to make sure replay handles more loading cases correctly.
3) Ghost emulator was triggering scroll events that were impossible to reach. This PR ensures we don't "wait" to scroll past the bottom of the page

Finally, we've removed the "UpstreamProxy" static callback, and simply made proxy configuration a setting per Secret Agent client that's instantiated. It seems to simplify the options, and since a session should emulate a single user session, the need to rotate IPs midway through a session seems to be an edge case we can revisit in future as need be. (Closes #92).
 - As part of this fix, enabled Socks5 proxies on top of http proxies (which were already supported)


client/index.ts
```diff
@@ -211,19 +212,19 @@ export function SecretAgentClientGenerator(
 
     /////// METHODS THAT DELEGATE TO ACTIVE TAB //////////////////////////////////////////////////////////////////////////
 
-    public async goto(href: string) {
+    public goto(href: string) {
```
calebjclark, line 215: Why remove `async`? I get that it's not "needed" but it's kinda nice for the code to signify what it returns. Should we give it an explicit return type?

core-interfaces/ICreateSessionOptions.ts
```diff
@@ -11,4 +12,5 @@ export default interface ICreateSessionOptions extends ISessionOptions {
   viewport?: IViewport;
   timezoneId?: string;
   locale?: string;
+  upstreamProxy?: IUpstreamProxy;
```
calebjclark, line 15: should this be `upstreamProxyUrl`?

calebjclark: Oh, never mind... sorry... I see 🙃

calebjclark: Well, opening this line of questing back up... do you think it would be simpler to just specify a `upstreamProxyURL` and let the `auth` (if needed) be in the URL?

website/docs/BasicInterfaces/SecretAgent.md
```diff
@@ -52,6 +52,7 @@ const SecretAgent = require('secret-agent');
   - renderingOptions `string[]`. Controls browser functionality.
   - userProfile `IUserProfile`. Previous user's cookies, session, etc.
   - showReplay `boolean`. Whether or not to show the Replay UI. Can also be set with an env variable: `SA_SHOW_REPLAY=true`.
+  - upstreamProxy `IUpstreamProxy`. A socks5 or http proxy url (and optional username) to use for all HTTP requests in this session. Dns over Tls requests will also use this proxy, if provided. Auth should be formatted as `user:password`. Http proxy auth will be converted to base64 basic authentication.
```
calebjclark, line 55: When it says, "and optional username" does that mean "and optional auth"? Or is it referring to username being in the url?

client/index.ts
blakebyrnes: Async without the await makes this kill off any further upstream stack trace. This is what I was telling you about that was behaving very strangely. I personally don't like cluttering the code with return types, but I can definitely see the argument for it, and maybe I'm just ok with it because the IDE is figuring it out from the ts compiler. I think if we decide to do that, we should probably make a global eslint rule requiring return types.

core-interfaces/ICreateSessionOptions.ts
blakebyrnes: The proxy configurations on the vpn providers tend to have separated url and auth, so I just thought this was a bit easier to translate without having to go make sure you've got formatting correct. This approach allows both. I previously set it up as separate properties (not an object), but then through there might be situations where want want additional configs on the proxy - like, bypassForDnsLookups, for an example. Does this lean you either way?

website/docs/BasicInterfaces/SecretAgent.md
blakebyrnes: Should say auth. The proxy configurations/docs will sometimes refer to this as username. I'm mixing terms I shouldn't be. Will correct

calebjclark:
What about just doing an await on the return?

client/index.ts
blakebyrnes: I added a rule to *-client, *-interfaces and commons to require return types on functions. I think this is a better direction for our code, but don't want to do it for all the code in one shot.

calebjclark: Great

core-interfaces/ICreateSessionOptions.ts
calebjclark: What if the URL includes auth details + auth key/value is include?

calebjclark: A simple upstreamProxyUrl just seems "simpler" than an object... I guess it seems that way to me since Chrome accept a string...

blakebyrnes: I was trying to make things easier, but if it's not easier, totally fine dropping it. The code will take auth precedence from the "auth" field over the url, but will take from either.

blakebyrnes: Removed.

Issue #98, opened 2020-12-04
UnhandledErrorOrRejection when trying to connect to unreachable proxy


xTRiM:
When proxy designated in upstreamProxyUrl is unreachable for whatever reason - process dies instead of throwing error.

**How to reproduce:**
```
const SecretAgent = require('secret-agent');

(async () => {
    try {
        const agent = await new SecretAgent({
            showReplay: false,
            upstreamProxyUrl: 'http://127.0.0.1:65535'
        });
        await agent.goto('http://api.ipify.org/');
    }
    catch(err) {
        console.error(`Error: ${err}`)
    }
})();
```
## Expected Behavior
Catchable error?

## Current Behavior
Getting a couple of errors in the console log.

```
ts-node ./src/_playground.ts

2020-12-04T10:53:48.829Z ERROR [@secret-agent/core/index] UnhandledErrorOrRejection {
  clientError: '2020/12/04 13:53:48 Dial (proxy/remote) Error: dial tcp 127.0.0.1:65535: connect: connection refused',
  context: {}
}
2020-12-04T10:53:48.831Z ERROR [@secret-agent/core/index] UnhandledErrorOrRejection {
  clientError: "TypeError: Cannot create property 'stack' on string '2020/12/04 13:53:48 Dial (proxy/remote) Error: dial tcp 127.0.0.1:65535: connect: connection refused'",
  context: {}
}
```
## Context
**SecretAgent version:** 1.2.0-alpha.2
**ts-node:** v9.1.0

PR #99, opened 2020-12-04
fixes for: Replay Multi-session, Mitm Socket, Proxy Errors

blakebyrnes:
This PR fixes 3 bugs:
1 - handle named sessions in replay (and don't overwrite the currently open script instance when multiple run at once)
2 - #98 properly bubble up proxy errors (and improve naming)
3 - mitm incorrectly tracking proxy connection back to itself, which means resources in concurrent sessions can get improperly assigned to other sessions

chore: require await when using async via eslint rule. Without this rule, we are seeing promise chains explode before reaching the client.


.eslintrc.js
```diff
@@ -172,6 +184,10 @@ module.exports = {
     '@typescript-eslint/no-empty-interface': 'off',
     '@typescript-eslint/no-namespace': 'off',
     '@typescript-eslint/ordered-imports': 'off',
+    '@typescript-eslint/no-shadow': [
+      'error',
+      { ignoreTypeValueShadow: true, ignoreFunctionTypeParameterNameValueShadow: true },
+    ],
```
calebjclark, line 190: Why?

.github/workflows/lint-and-test.yml
```diff
@@ -116,7 +116,7 @@ jobs:
           SA_REPLAY_SKIP_BINARY_DOWNLOAD: 1
 
       - name: Run tests
-        run: yarn jest --testTimeout=15000 --runInBand
+        run: yarn jest --testTimeout=60000 --runInBand
```
calebjclark, line 119: That's a major increase.

.prettierrc
```diff
@@ -2,5 +2,6 @@
   "printWidth": 100,
   "singleQuote": true,
   "trailingComma": "all",
-  "tabWidth": 2
+  "tabWidth": 2,
+  "arrowParens": "avoid"
```
calebjclark, line 6: Why? https://prettier.io/docs/en/options.html#arrow-function-parentheses

blakebyrnes: This was causing changes to the code base as it stands currently everytime I was hitting save and auto-formatting. I chose to change the rule to match the current code

.github/workflows/lint-and-test.yml
blakebyrnes: It was an attempt to workaround these sporadically failing tests that happens almost exclusively on windows. I didn't spend the time to figure out how to change for only windows.

.eslintrc.js
blakebyrnes: Cause you love shadows?!?! I would have just made it a straight error, but ... you love shadows :)

blakebyrnes: I was trying to find the equivalent rule to our old rule about shadows being ok when they don't change the outer scope, but this was the closest I could get with the @typescript-eslint module

PR #104, opened 2020-12-11
Handlers, default agent, native remote

blakebyrnes:
This is a very large change. I'd like to get your thoughts pulling down the branch. It's technically "done" and "working", but I marked as WIP so we can talk through it first.

I plan on writing a blog post explaining why we did this that we can put on the site and also potentially in the "discussions" page on github.

Changes in the PR:
- Client merges in remote client and now operates more like how Remote Client was working before
- Full client exports a "Default" agent that is ready to use instantly hooked up to a local Core.
- Remote clients are natively supported and you simply need to launch a server where your core will run.
- Handlers: handlers provide a concurrency mechanism for working with multiple remote or local cores and generating many Agents.
- Core now creates "connections" with each client, which means we track sessions a bit differently, and the way we route commands changed a bit.
- Client has added notion of "sessions" and stopped operating on an independent-tab basis, which is what it was doing previously.

Things I could use help on (that we haven't talked through verbatim already):
1) Check the docs. Does the lingo we're using make sense?
2) Check the code for handlers, remote tests, etc to make sure they feel good
3) Check core/index and core/lib/CoreServerConnection - this is the part I feel least good about right now, but wondering if it's good enough for now since it's not really impacting client side?


full-client/index.ts
```diff
@@ -1,34 +1,39 @@
...
+    coreServerConnection.on('message', payload => this.onMessage(payload));
   }
-  return { data: await Core[command](...args) };
-};
 
-// INCOMING ////////////////////////////////////////////////////////////////////
+  sendRequest(payload: ICoreRequestPayload): void | Promise<void> {
+    return this.coreServerConnection.handleRequest(payload);
+  }
+}
```
calebjclark, line 24: This seems to want it's own file. lib/LocalCoreConnection.ts?

calebjclark: Why not just put it in client/lib alongside RemoteCoreConnection.ts? The class itself doesn't need access to Core since it accepts coreServerConnection instance.

full-client/index.ts
```diff
@@ -1,34 +1,39 @@
...
 
-Core.onEventFn = (meta: ISessionMeta, listenerId: string, ...args: any[]) => {
-  coreClient.pipeIncomingEvent(meta, listenerId, args);
+CoreClient.LocalCoreConnectionCreator = (options: ICoreConnectionOptions) => {
+  const coreServerConnection = Core.addConnection();
+  return new LocalCoreConnection(options, coreServerConnection);
 };
```
calebjclark, line 29: I really like how you hooked this in from the `client` pkg. Very clean separation.

website/docs/BasicInterfaces/Agent.md
```diff
@@ -1,58 +1,45 @@
...
-const SecretAgent = require('secret-agent');
+const agent = require('secret-agent');
 
 (async () => {
-  const agent = new SecretAgent();
   await agent.goto('https://www.google.com');
   // other actions...
   await agent.close();
 })();
 ```
```
calebjclark, line 14: Jump into the "Agent class" through me for a loop when instead of a class example I was shown an instance... I'm not sure what to do here, but I find it difficult to make the jump.

website/docs/BasicInterfaces/Agent.md
```diff
@@ -1,58 +1,45 @@
...
-  const agent = await new SecretAgent();
+  const handler = new Handler();
+  handler.dispatchAgent(async agent => {
+    await agent.goto('https://www.google.com/search?q=secret-agent');
+  });
+  handler.dispatchAgent(async agent => {
+    await agent.goto('https://www.bing.com/search?q=secret+agent');
+  });
+  await handler.waitForAllDispatches();
 })();
```
calebjclark, line 33: I'm not sure we want to give a Handler example... linking to Handler is probably enough. Diving into an example that's not relevant to the topic at hand makes the page too complex.

website/docs/BasicInterfaces/Agent.md
calebjclark: I wonder if we need to export the Agent class so it's clear that lowercase `agent` is an instance of Agent and not just a singleton that we lowercased.

website/docs/Advanced/Remote.md
```diff
@@ -1,56 +1,37 @@
...
+(async () => {
+  const server = new SecretAgentServer();
+  await server.listen(7007);
+})().catch(console.log);
+```
 
 ## Setting Up the Client
 
-Below is a client example that uses node's net socket library.
+Your [Handler](../basic-interfaces/handler) will be configured to point at this Remote Core (and any others you've set up).
```
calebjclark, line 23: Slight word change... "Your Handler MUST be configured..."

website/docs/Advanced/Remote.md
```diff
@@ -1,56 +1,37 @@
...
+
+NOTE: you can install the `@secret-agent/client` npm package if you don't want to install a full browser engine on the machine coordinating all your scrapes. That example is shown below.
```
calebjclark, line 25: "You can USE the..."

full-client/index.ts
blakebyrnes: CoreServerConnection is part of core, so it can't go in client, but we can definitely move it to it's own class. I had previously created a small interface that LocalCoreConnection took in, which I can do again, but I kind of liked seeing it wired up in the same file as the "LocalCoreConnectionCreator" part of the file so you could understand how it was getting wired up. I'm kind of cool with either path, but I tried several options here, and having it sit in the "middle or core and client" seemed to make the most sense - it feels like the many-to-many class.

website/docs/BasicInterfaces/Agent.md
blakebyrnes: The example you've highlighted has the agent being exported from the secret-agent module right above. Did you miss that part?

I agree it's a bit jarring, but I think you come into the website if you start here, but I think you'll start on the Overview page, not here.

website/docs/BasicInterfaces/Agent.md
blakebyrnes: I started this off with a mention of Handler, and it felt like I was not doing justice to explaining what all you can do. I think your thought is very valid though. An "Agent" comes from 2 places - a handler, or a default export. This "file" explains how to use it, so it feels very relevant to explain how you get one. Since this "class" is no longer "constructed", it seemed like we need to be showing you how that happens.

full-client/index.ts
calebjclark: Your logic makes sense. 

I guess it just felt/feels weird to me that of our two CoreConnection classes, they're split across two different packages. It seems they should be together (along with UlixeeCoreConnection when we add that).

There's nothing else in full-client that defines new functionality or classes... it's just some dependencies and glue code.

blakebyrnes: I guess I can see the appeal of having them near each other. However, they each have separate, optional dependencies. If we abstract the LocalCoreConnection so it takes in an interface, would you be in favor of a "connections" folder in client, so we don't have to flip the naming to the end to keep them in the same place? Or should I flip the names on all of them - CoreClientConnection, CoreConnectionRemote, CoreConnectionLocal, etc?

website/docs/BasicInterfaces/Agent.md
calebjclark: I think you mean imported. Yes, I see that it's imported from the `secret-agent` library. That doesn't change anything for me.

I know we both agreed only the `agent` instance and Handler class should be exported from `secret-agent` (not Agent), but seeing the documentation is making me question that decision.

We've hidden some of the dots connecting how everything ties together. Yes, it's jarring, but it also just doesn't quite make sense.

I've been going back and forth on what to do about this. I wonder if we should just add a new "Default Export" page as the first item in Basic Interfaces.

full-client/index.ts
calebjclark: Yes

blakebyrnes: This has been updated, as well as unifying full-client and client exported classes/instances, and ability to set a CoreConnection on an agent

blakebyrnes:
I've changed the structure based on some conversations with @calebjclark and things feel far simpler. Agent constructor is re-documented and takes a coreConnection if desired, so you can easily connect to remote Cores. As part of this change CoreClient is split into CoreClientConnection and CoreSessions (which tracks the open sessions for a connection and manages concurrency).

Also as a part of this push, I added a blog post about Handlers. @calebjclark if you can boot up the website, let me know what you think.

website/blog/2020-12-17-why-handlers.md
```diff
@@ -0,0 +1,122 @@
...
+
+  await agent.close();
+})();
+```
+
+#### Step 2: Run Multiple Scrapes
+
+Agent instances are lightweight, but what do you do when you need to queue up thousands of them to run. Until now, you've been on your own to use libraries like `p-queue`, keeping track of promises, or simply waiting and looping.
+
+We've introduced a new idea into SecretAgent called a [`Handler`](/docs/basic-interfaces/handler) to help run multiple scrapes in one session. Handlers manage the concurrency of multiple scrapes to ensure your machine doesn't get overloaded and hang. We've designed it so that your code should require almost no changes to transition to many scrapes.
```
calebjclark, line 45: "We designed it..."

website/blog/2020-12-17-why-handlers.md
```diff
@@ -0,0 +1,122 @@
...
+  handler.dispatchAgent(async agent => {
+    // agent is automatically created for us
+    await agent.goto('https://ulixee.org');
+    ...
+```
+
+
+#### Default Exports
+
+To get to this setup, you'll notice some changes in the default exports when you install SecretAgent 1.3.0. The default exports that come out of the `secret-agent` and `@secret-agent/client` packages is now a pre-initialized instance of the `Agent` class (`SecretAgent` was renamed to `Agent`).
```
calebjclark, line 115: 1.3.0-alpha

website/blog/2020-12-29-why-handlers.md
```diff
@@ -0,0 +1,122 @@
...
+- Do you create a new SecretAgent instance every time? Or do you simply add tabs?
+- How expensive is it to create many instances?
+- How should I make sure not to overload the host machine with the number of scrapes running at the same time?
+- How do I add new machines when I max out the current one?
+
+As we've explored simplifying this story, we want to make the progression of "examples" through to full-scrapes a smooth process. Something like this:
+
+#### Step 1: Try Out an Example
+
+Trying out examples should require as little setup as possible, so we've added a new `default export` that's a ready-to-go client for SecretAgent.
```
calebjclark, line 21: "so we added..."

website/blog/2020-12-29-why-handlers.md
```diff
@@ -0,0 +1,122 @@
+---
+title: 'Scaling SecretAgent Scrapes with Handlers'
+path: /handling-scale
+date: 2020-12-29
+summary: "We needed a simpler approach to scaling out to multiple machines running SecretAgent and 1000s of waiting actions. So we've added a new concept called Handlers."
+---
+
+When you start using SecretAgent, you often copy and paste the default examples. As we've started to use SecretAgent on larger extraction efforts, it's become clear that we don't have a clear story for "how" you go from that starting example to running 2, or even 1000 scrapes.
```
calebjclark, line 8: "As we started..."

calebjclark: "it became clear..."

PR #110, opened 2020-12-22
Fix typo

SneakyFish5:
I think this is a typo? Sorry if I'm mistaken


calebjclark:
Good catch @SneakyFish5. It was definitely a typo, but no longer.... thanks for the pull request!

PR #116, opened 2020-12-29
feat: rename renderingOptions => blockedResourceTypes

### LINKED ISSUE #113: How to block certain types of HTTP Requests from loading?
I want to block HTTP Requests that doesn't affect scrapping process like images, styles, etc. How can I do it? Does it increases the chances of detection of Secret Agent?

blakebyrnes:
BREAKING CHANGE: renames “renderingOptions” to “blockedResourceTypes”. Default is now “None”

Closes #113

### DIFF
--- a/full-client/test/basic.test.ts
+++ b/full-client/test/basic.test.ts
@@ -1,4 +1,5 @@
 import { Helpers } from '@secret-agent/testing';
+import Resource from '@secret-agent/client/lib/Resource';
 import { Handler } from '../index';
 
 let koaServer;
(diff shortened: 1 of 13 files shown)


full-client/test/basic.test.ts
```diff
@@ -1,4 +1,6 @@
 import { Helpers } from '@secret-agent/testing';
+import ResourceType from '@secret-agent/core-interfaces/ResourceType';
```
calebjclark, line 2: I thought our linter used to catch these?

calebjclark: Oh, it is... ha. Should we fix the 3 lint issues in PR?

blakebyrnes: It did... warning not error. Will fix.

PR #121, opened 2020-12-30
fix(core): elements offscreen fail isVisible

blakebyrnes:
Very small fix to make an element that is offscreen count as "not-visible"


calebjclark:
It's hard to make a smaller PR than this one 🙃

blakebyrnes:
I like to think of it as a sign of maturity .. ha

PR #124, opened 2021-01-01
Added more emulators

calebjclark:
Chrome81
Chrome84
Chrome85
Chrome86
Chrome87


blakebyrnes (approved):
The changes look good. Main questions/comments are:
1 - We don't seem to have added any testing for the new engines. Should we?
2 - We added a "marketshare" to config.json - do we need an as-of date? Maybe not since I'm not sure what we could do with the date anyway...
3 - this doesn't seem to have changed the default emulator to chrome 87. Should it? Or do we need more testing first?

PR #130, opened 2021-01-07
Convert RemoteServer/Client and Replay API to use Web Sockets

blakebyrnes:
This PR moves the Core Api and Replay Api to use websockets instead of http2/json-socket.


replay/backend/api/index.ts
```diff
@@ -1,7 +1,7 @@
-import * as http2 from 'http2';
-import { ClientHttp2Stream, IncomingHttpHeaders, IncomingHttpStatusHeader } from 'http2';
+import * as WebSocket from 'ws';
 import { ChildProcess, spawn } from 'child_process';
 import * as Path from 'path';
+import { IncomingHttpHeaders } from 'http';
```
calebjclark, line 4: defined but never used

PR #131, opened 2021-01-08
fix(puppet): launch non-default (83) headed chrome

blakebyrnes:
Fixes issue where multiple browsers opening headed will close the first one, which shuts everything down.

### DIFF
--- a/puppet-chrome/index.ts
+++ b/puppet-chrome/index.ts
@@ -1,16 +1,27 @@
-import IPuppetBrowser from "@secret-agent/puppet-interfaces/IPuppetBrowser";
-import ILaunchedProcess from "@secret-agent/puppet-interfaces/ILaunchedProcess";
-import IPuppetLauncher from "@secret-agent/puppet-interfaces/IPuppetLauncher";
-import os from "os";
-import { Browser } from "./lib/Browser";
-import { Connection } from "./lib/Connection";
+import IPuppetBrowser from '@secret-agent/puppet-interfaces/IPuppetBrowser';
+import ILaunchedProcess from '@secret-agent/puppet-interfaces/ILaunchedProcess';
+import IPuppetLauncher from '@secret-agent/puppet-interfaces/IPuppetLauncher';
+import os from 'os';
+import Path from 'path';
+import { Browser } from './lib/Browser';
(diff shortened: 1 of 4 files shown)


puppet-chrome/index.ts
```diff
@@ -1,16 +1,22 @@
...
-import ILaunchedProcess from "@secret-agent/puppet-interfaces/ILaunchedProcess";
-import IPuppetLauncher from "@secret-agent/puppet-interfaces/IPuppetLauncher";
-import os from "os";
-import { Browser } from "./lib/Browser";
-import { Connection } from "./lib/Connection";
+import IPuppetBrowser from '@secret-agent/puppet-interfaces/IPuppetBrowser';
+import ILaunchedProcess from '@secret-agent/puppet-interfaces/ILaunchedProcess';
+import IPuppetLauncher from '@secret-agent/puppet-interfaces/IPuppetLauncher';
+import os from 'os';
+import Path from 'path';
```
calebjclark, line 5: Shouldn't this be `import * as Path from 'path'`?

puppet-chrome/index.ts
```diff
@@ -82,8 +90,6 @@ const defaultArgs = [
   '--allow-running-insecure-content',
 
   '--lang=en-US,en;q=0.9',
-  '--window-size=1920,1080',
-  '--window-position=0,0',
```
calebjclark, line 86: Why did you get rid of these?

blakebyrnes: I figured it was kind of weird to set those there since we're setting screensize in code now for every request.

puppet-chrome/index.ts
blakebyrnes: Only in strict mode, which we don't have turned on in SA.

calebjclark: What standard should we follow in our code? It seems we have it both ways in the code base.

puppet-chrome/index.ts
calebjclark: Agreed

puppet-chrome/index.ts
blakebyrnes: Good question. I didn't realize how split it was.

PR #132, opened 2021-01-08
Single Replay, Shared Directory withe Browsers, Better Launch errors

blakebyrnes:
This PR addresses a few things with our external dependencies:
*Replay*
- Replay now has single install for all core versions in a project (the newest will be installed)
- Replay lives in the same directory as the Chrome engines
- Replay locks on files to create a path to an api server where scripts register. This prevents the electron bug for popping up so many processes to try to determine if a new one should launch. Scripts now register over http calls instead

*Puppet:*
- Clean up sandbox errors (copied from playwright)
- Better error handling if a chrome version is not installed (previously was showing an ENOENT error without explicitly saying what was happening)
- Puppet engines share a common directory inside app cache (secret-agent)

### DIFF
--- a/puppet-chrome/index.ts
+++ b/puppet-chrome/index.ts
@@ -54,6 +54,27 @@ const PuppetLauncher: IPuppetLauncher = {
       throw error;
     }
   },
+  translateLaunchError(error: Error): Error {
+    // These error messages are taken from Chromium source code as of July, 2020:
+    // https://github.com/chromium/chromium/blob/70565f67e79f79e17663ad1337dc6e63ee207ce9/content/browser/zygote_host/zygote_host_impl_linux.cc
+    if (
+      !error.message.includes('crbug.com/357670') &&
+      !error.message.includes('No usable sandbox!') &&
+      !error.message.includes('crbug.com/638180')
+    ) {
+      return error;
+    }
+    error.stack += [
+      `\nChromium sandboxing failed!`,
+      `================================`,
+      `To workaround sandboxing issues, do either of the following:`,
+      `  - (preferred): Configure environment to support sandboxing (eg: in Docker, use custom seccomp profile + non-root user + --ipc=host)`,
+      `  - (alternative): Launch Chromium without sandbox using 'chromiumSandbox: false' option`,
+      `================================`,
+      ``,
+    ].join('\n');
+    return error;
+  },
 };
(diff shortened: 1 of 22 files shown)


puppet-chrome/index.ts
```diff
@@ -54,6 +54,27 @@ const PuppetLauncher: IPuppetLauncher = {
...
+      !error.message.includes('crbug.com/638180')
+    ) {
+      return error;
+    }
+    error.stack += [
+      `\nChromium sandboxing failed!`,
+      `================================`,
+      `To workaround sandboxing issues, do either of the following:`,
+      `  - (preferred): Configure environment to support sandboxing (eg: in Docker, use custom seccomp profile + non-root user + --ipc=host)`,
+      `  - (alternative): Launch Chromium without sandbox using 'chromiumSandbox: false' option`,
```
calebjclark, line 72: I wonder if we should put up a doc page on exactly how to do this. It's not immediately clear on exactly how I would do this.

calebjclark: It's especially unclear if this gets thrown while I'm using SA.

blakebyrnes: I wasn't sure how far to go in copying these in from Playwright and Puppeteer. Playwright does a great job actually inspecting the runtime shared libraries that are installed, but they do it by distributing a custom windows executable. I got a bit stalled out figuring out what was right to bring in right now. FWIW, here are some references from the two platforms:

- Playwright docker guide: https://github.com/microsoft/playwright/blob/master/docs/src/docker.md
  - NOTE: they're publishing multiple dockers, a seccomp profile. I think we would just want the section on "Run the image" if we go this path, assuming this seccomp is desirable outside of their setup. We would also probably need to host the modified seccomp.json profile in our project just like playwright does.
- Puppeteer: https://github.com/puppeteer/puppeteer/blob/main/docs/troubleshooting.md
  - NOTE: they're recommending a different approach for setuid..

So anyway... I am not sure if we should pick our own path right now? Recommend Playwright's path (since it sounds a little more up to date) or what... I think we could probably use several sections of Puppeteer's troubleshooting guide too. Is this something we should do now? Or just link to them?

calebjclark: I'm not sure the right approach, so perhaps we should punt for now.

PR #134, opened 2021-01-12
fix(emulators): safari cookie handling fix

blakebyrnes:
Safari has specific handling around cookies that didn't really have anything exercising page triggered cookies. Tweaked the test to make sure we hit this code branch, and fixed the error.

NOTE: this has a todo for you @calebjclark to inject either a sessionid or a logger to the emulators. I thought doing too much in here in this PR might break your stuff.

### DIFF
--- a/emulate-browsers/safari-13/index.ts
+++ b/emulate-browsers/safari-13/index.ts
@@ -24,7 +24,8 @@ import IHttpResourceLoadDetails from '@secret-agent/core-interfaces/IHttpResourc
 import IResolvablePromise from '@secret-agent/core-interfaces/IResolvablePromise';
 import { createPromise, pickRandom } from '@secret-agent/commons/utils';
 import IUserProfile from '@secret-agent/core-interfaces/IUserProfile';
-import IWindowFraming from "@secret-agent/core-interfaces/IWindowFraming";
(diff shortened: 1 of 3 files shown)


emulate-browsers/safari-13/index.ts
```diff
@@ -135,7 +137,14 @@ export default class Safari13 {
     const result = this.domOverrides.build();
     Object.assign(result[0], {
       callback: ({ cookie, origin }) => {
-        this.cookieJar.setCookie(cookie, origin);
+        try {
+          this.cookieJar.setCookie(cookie, origin);
+        } catch (error) {
+          log.warn('Error setting cookie from page', {
+            error,
+            sessionId: null, // TODO: @caleb - can you plug in sessionid to the emulators (or a child logger)
```
blakebyrnes, line 145: left a @calebjclark todo for the next pr

calebjclark: Where would I plug sessionId into the other emulators? There are no similar callbacks in the other emulator scripts.

blakebyrnes: You would set the logger/sessionid on the emulator when it's constructed. So we'd be adding a property that all emulators would have injected

PR #150, opened 2021-02-05
Updated emulators

calebjclark:
- Chrome emulators are now being dynamically generated from Slab. 
- Safari12 emulator uses sessionId
- Big Sur data is now included in all emulators
- Emulator-base has better OS support, especially for OS versions which it doesn't recognize... it falls back to closest.
- All data within emulator-base is now pulled from Slab.


core/lib/Session.ts
```diff
@@ -80,6 +80,7 @@ export default class Session extends TypedEventEmitter<{
     if (options.userProfile) {
       this.userProfile = options.userProfile;
       this.browserEmulator.userProfile = options.userProfile;
+      this.browserEmulator.sessionId = this.id;
```
blakebyrnes, line 83: I think we likely want to set sessionId even if there's not a userProfile, don't we?

calebjclark: Agreed. Pushed fix.

Issue #159, opened 2021-02-10
Core export problem, possible issue with ECONNREFUSED


A-Posthuman:
While experimenting with the "server" part of the Remote docs, running into the issue that Core is requiring adding .default to the require line. Also in that example the log.info lines had to be swapped to console.log.

After then getting the code ready to test on client/server, accidentally tried to first start the "server", which then immediately failed to be able to connect to client, leading to endless huge stream of:

> 2021-02-10T04:52:52.267Z ERROR [/home/ubuntu/node_modules/@secret-agent/client/lib/Handler] UnhandledRejection {
>   error: 'Error: connect ECONNREFUSED

I found I can catch this error by wrapping my agent.goto with try/catch, but wasn't sure if this is the desired way to go. I guess the cleaner way to go would be to use RemoteConnectionToCore and the .connect() method to verify I have an ok connection first. So I've modified my program to try that with some code like this:

```
process.env.SA_SHOW_REPLAY="false";
const { Handler, RemoteConnectionToCore } = require('secret-agent');
async function run() {
    try {
        var remote1 = new RemoteConnectionToCore({ maxConcurrency: 2, host: '172.31.30.34:2600' });
        await remote1.connect();
    } catch (err) {
        console.log(`Caught an exception: ${err}`);
        console.log('Problem setting up RemoteConnectionToCore, exiting.');
        process.exit();
    }
    
    try {
        var handler = new Handler({ remote1 });
    } catch (err) {
        console.log(`Caught an exception: ${err}`);
        console.log('Problem setting up Handler, exiting.');
        process.exit();
    }

    handler.dispatchAgent(async agent => {
        // agent is automatically created for us
        try {
            await agent.goto('https://ulixee.org');
        } catch (err) {
            console.log(`Caught an exception: ${err}`);
            console.log('Problem with agent.goto, exiting.');
            process.exit();
        }

more stuff...

}
run();
```

However I'm catching an error where I try to do the first goto with the new Handler:

> 2021-02-10T05:34:20.975Z ERROR [/home/ubuntu/node_modules/@secret-agent/client/c
> onnections/ConnectionFactory] Error connecting to core {
>   error: 'TypeError: Converting circular structure to JSON\n' +
>     "    --> starting at object with constructor 'RemoteConnectionToCore'\n" +
>     "    |     property 'commandQueue' -> object with constructor 'CoreCommandQu
> eue'\n" +
>     "    --- property 'connection' closes the circle",
>   context: {},
[...]

Issue #158, opened 2021-02-10
Console logger seems to ignore process.env.NODE_DISABLE_COLORS = 1


A-Posthuman:
I was attempting to remove the color codes when dumping some error and debug outputs to text file, but the process.env.NODE_DISABLE_COLORS = 1 option I googled doesn't seem to do the trick. Would it be possible to update your console logger to look for this?

A-Posthuman:
Also added a couple of potential issues onto closed Issue #153

blakebyrnes:
I'm not sure where those would be coming from. If you look in commons/Logger.ts - that's where the logging is happening. There's nothing else being applied to the logs. Perhaps something in the terminal program you're using?

For what it's worth, all logs are in the Session database for each run in SessionLogs (it's a sqlite3 file). You don't need to run DEBUG=true to get those to be captured. It might be what you're looking for...

Thanks for all the issue reports!

A-Posthuman:
I'm not sure, but do the Sessions also capture all the stack traces from errors that halt the program?

I believe the coloring of some of the error messages may come from within Node itself, for example using this simple program to generate an error:

```
// process.env.NODE_DISABLE_COLORS = 1;
const test = "test";
test = "error";
```

Results in my PuTTY getting colored text (some of it grey) like this:

![Untitled](https://user-images.githubusercontent.com/78522671/107601061-34aa6180-6beb-11eb-809b-69d2da2ce60c.png)

If I want to redirect the program outputs to a log file or something to look at later, it ends up having a lot of ugly ANSI color codes in it unless I pipe it all through a sed command to strip those.

So if I uncomment that process.env.NODE_DISABLE_COLORS line, then in that program it does revert to not using colors:

![Untitled](https://user-images.githubusercontent.com/78522671/107601174-8ce16380-6beb-11eb-95c9-470705d820b1.png)

However when I try to use that same line to disable colored output in secret-agent, I still get colored output like this:

![Untitled](https://user-images.githubusercontent.com/78522671/107601260-cca84b00-6beb-11eb-8eb2-7e7bcda918fc.png)

So somehow my ENV variable is being ignored or overridden somewhere?

blakebyrnes:
Oh, I see where we are actually using colors. I'll put a check for the disable colors in here.

Issue #153, opened 2021-02-07
Some unexpected errors from repetitively goto'ing same URL


A-Posthuman:
While testing with a simple loop that repetitively requests the same URL, I've run into a some different unexpected error situations.

In one case, requesting https://www.whatismyip.com/ led to some uncaught DNS-related errors. Sample code & results txt:

```
process.env.SA_SHOW_REPLAY="false";
const { Agent } = require('secret-agent');

async function run() {
  var agent = await new Agent();
  let url = 'https://www.whatismyip.com/';
  for (let i = 0; i < 10; i++) {
    try {
      console.log('here 1');
      let resource = await agent.goto(url, 5000);
      console.log('here 2');
      let statusCode = await resource.response.statusCode;
      console.log(`here 3: ${statusCode}`);
    } catch (err) {
      console.log(`Caught an exception: ${err}`);
    }
  }
  await agent.close();
}
run();
```

Results:
[Results1.txt](https://github.com/ulixee/secret-agent/files/5938243/Results1.txt)

I'm not clear where the reference to ipv6.whatismyip.com came from, does that site sometimes redirect to that as an anti-bot technique? Also not clear if this is the expected behavior from secret-agent for a DNS lookup error, in terms of the unhandled errors?

In a 2nd case, repeatedly requesting https://www.slashdot.org/ using the same loop code, led to:

Results:
[Results2.txt](https://github.com/ulixee/secret-agent/files/5938246/Results2.txt)

Not clear what caused that or if it is important.

Lastly, repeatedly requesting http://www.posthuman.com/ leads to the requests timing out after the 1st success. I've tried removing my 5 sec timeout and then the normal 30 sec timeout eventually occurs. This site is a basic web hosting site on 1and1/Ionos shared hosting. They advertise as providing anti-DDOS protection, could this behavior be some anti-bot detection system, or is it something else? The behavior is perfectly repeatable, always after starting the program it gets 1 working 200 status response, then everything times out after that.

Results:
[Results3.txt](https://github.com/ulixee/secret-agent/files/5938247/Results3.txt)

Results without me catching the error:
[Results4.txt](https://github.com/ulixee/secret-agent/files/5939359/Results4.txt)

blakebyrnes:
@A-Posthuman thanks for reporting! 

[...]

Issue #108, opened 2020-12-18
Fails creepjs tests


felyxjet:
[creepjs](https://abrahamjuliot.github.io/creepjs/) is able to detect secret-agent's lies.

blakebyrnes:
@felyxjet Nice profile picture ;) 

Thanks for the report. Looks like a lot of great tests in creepjs that we haven't gotten to in our Double Agent testing quite yet. In particular, we're failing a lot of tests that originate in Service Workers. Will make that a priority in our upcoming additions.

felyxjet:
Nice, looking forward to it. 

P.S. it's not a surprise i like your profile picture as well ))

PR #161, opened 2021-02-11
Worker UserAgents/Platform, Bug fixes

blakebyrnes:
This PR handles appropriately setting useragent and platform in workers. Also ensures on a newly navigated page, the platform and useragent match the emulator.

Misc Fixes
- Don't use colors in console if turned off in node env variables. Linked to #158 
- Update electron version (security warning)
- Cleanup variables sent to CoreServer during a connect. Was serializing anything, which could lead to loops if a connection was passed in. Linked to #159 
- Return information to client when responses bypass the Mitm and use browser caches #153 
- Fix some failing tests
- Fix toString leaving evidence of proxy #108

### DIFF
--- a/emulate-browsers/chrome-80/index.ts
+++ b/emulate-browsers/chrome-80/index.ts
@@ -14,7 +14,7 @@ import IUserAgentOption from '@secret-agent/core-interfaces/IUserAgentOption';
 import { randomBytes } from 'crypto';
 import IUserProfile from '@secret-agent/core-interfaces/IUserProfile';
 import { pickRandom } from '@secret-agent/commons/utils';
-import IWindowFraming from "@secret-agent/core-interfaces/IWindowFraming";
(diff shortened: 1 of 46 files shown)


emulate-browsers/chrome-80/index.ts
```diff
@@ -104,6 +104,22 @@ export default class Chrome80 {
...
+      'MediaDevices.prototype.enumerateDevices',
+      'Notification.permission',
+      'Permission.prototype.query',
+      'WebGLRenderingContext.prototype.getParameter',
+      'HTMLMediaElement.prototype.canPlayType',
+      'RTCRtpSender.getCapabilities',
+    ]);
+    return result;
+  }
+
```
calebjclark, line 122: I'm not sure how we should handle changes to Emulators. This file is auto-generated by Slab. I'm not sure if it should be directly modified here.

Issue #163, opened 2021-02-13
Is there a way for an agent to determine what remote Handler host it's on?


A-Posthuman:
Is there a way via agent.meta or other means for my code to figure out which remote host connection/IP address it's getting a particular result from? There may be some cases I'm imagining where if I'm farming out Handler requests to several different hosts, and one host starts having a problem (say for example its IP address gets blocked from a site I'm scraping), I need to be able to tell which host has the issue, then somehow remove it from the connections the Handler is farming stuff to (I guess I'd have to end the Handler and make a new one?).

blakebyrnes:
I don't think we do have that at the moment, but this does make sense. @calebjclark should we call this property "coreHost" on Agent? Is there any reason you can think of we'd want to expose the whole connection object?

A-Posthuman:
I could see it might be useful if my Handler control code could be able to not only know which IP address/connection is causing a problem, but also be able to dynamically disconnect it from the Handler pool and have no more dispatches sent to the disconnected connection (and any that were internally queued or in progress with that connection be shifted to other connections... not sure how your internal queuing works). Does the ConnectionToCore disconnect() method do something like that?

blakebyrnes:
Good feedback and thought. So perhaps it makes more sense to expose coreConnection.

calebjclark:
@A-Posthuman are you planning to setup/connect to multiple hosts/Cores? I'm curious why, but I also imagine this architecture will impact how we should expose connection capabilities.

Are you thinking each host/Core will be using a single IP address? Have you looked at using a single Core with rotating proxies?

A-Posthuman:
I'm conducting experiments at the moment yes with each remote host having a single IP, and then instituting some type of monitoring of the results each host gets back, and if a host starts to get too much throttling/blockage then terminate that host and setup a fresh one to take its place with new IP.

Not sure if this is the architecture I'll end up using, but if I do it would be on the order of 20 to 100 remote hosts in the production system depending on what throughput I can achieve per host per hour and other factors.

Also not sure yet what are the inherent limits of your Handler dispatcher, in terms of how many remote hosts and queued work/dispatches it can handle in a given server size before hitting mem or cpu limits. I'll have to run some tests to see how far it can scale up.

[...]

PR #172, opened 2021-02-19
feat(client): coreHost & disconnecting errors

### LINKED ISSUE #165: Some errors using connection.disconnect(), client/server hang in some cases
Experimenting with setting up a Handler that has both a remote host connection and a local connection, then using setTimeout after 4000 ms to trigger disconnect() on the remote connection. If I don't have await handler.waitForAllDispatches() wrapped with try/catch then this tends to result in an error that is duplicated to both client/server consoles, but then plus some extra console logging starting with "clientError: 'Error: Canceling Queue Item'" on the "client":

[debug.txt](https://github.com/ulixee/secret-agent/files/5993161/debug.txt)

this leaves both the client & server in a hung state requiring control-c to end. I'd guess the client code is hung at await handler.waitForAllDispatches() because it doesn't get to my console print of 'Done waiting'...

On other runs of the same code I get somewhat different results, sometimes it seems the local connection will continue on and process the last few of the urls/costs, here is a console log of that run, however notice on this one it didn't seem to print my "Done with disconnect" from the end of my setTimeout function. Not sure if it got hung awaiting the remote1.disconnect() on the line before that? It hangs on this run too.

[debug2.txt](https://github.com/ulixee/secret-agent/files/5993200/debug2.txt)

Another iteration is if I uncomment the try/catch code around the await handler.waitForAllDispatches(), then I get 3 possible behaviors:

1) No error, the console "Problem with waitForAllDispatches" never prints, the local connection seems to finish the dispatches normally.

2) An error disconnecting, console prints "Problem with waitForAllDispatches" on client, however the remote server catches nothing/has no error and continues running, and in this iteration the local dispatches seem to stop usually and not complete getting all the costs.

3) Error: read ECONNRESET stack:

[debug4.txt](https://github.com/ulixee/secret-agent/files/5993244/debug4.txt)

Lastly of note, in the ------AGENT QUEUE--------------------------------- portion of the stack trace on the "client", and also in the last debug file with the 'Error: read ECONNRESET, it has that issue of ignoring my process.env.NODE_DISABLE_COLORS = 1 setting.

Here's the client code:

```
// process.env.DEBUG = true;

// Set this ENV variable to make sure secret-agent knows this is a headless server, so don't try to start Replay
process.env.SA_SHOW_REPLAY="false";
process.env.NODE_DISABLE_COLORS = 1;

const { Handler, RemoteConnectionToCore } = require('secret-agent');

async function run() {
    try {
        var remote1 = new RemoteConnectionToCore({ maxConcurrency: 2, host: '172.31.19.191:2600' });
        await remote1.connect();
    } catch (err) {
        console.log(`Caught an exception: ${err}`);
        console.log('Problem setting up RemoteConnectionToCore, exiting.');
        process.exit();
    }
    
    try {
        var handler = new Handler( remote1, { maxConcurrency: 1 } );
    } catch (err) {
        console.log(`Caught an exception: ${err}`);
        console.log('Problem setting up Handler, exiting.');
        process.exit();
    }

    handler.dispatchAgent(async agent => {
        // agent is automatically created for us
        try {
            await agent.goto('https://ulixee.org');
            const datasetLinks = await agent.document.querySelectorAll('a.DatasetSummary');
            for (const link of datasetLinks) {
                const name = await link.querySelector('.title').textContent;
                const href = await link.getAttribute('href');
                const dataset = { name, href };
                // add a name to each agent so we can find each scrape on Replay
                const agentOptions = { name };
                console.log(`name: ${name}, href: ${href}`);
                handler.dispatchAgent(getDatasetCost, dataset, agentOptions);
            }
        } catch (err) {
            console.log(`Caught an exception: ${err}`);
            console.log('Problem with dispatchAgent');
            process.exit();
        }
    });
    console.log('About to waitForAllDispatches');
    setTimeout( async function () {  
        console.log("setTimeout triggered");
        await remote1.disconnect();
        console.log("Done with disconnect");
    }, 4000);
    // try {
        await handler.waitForAllDispatches();
    // } catch (err) {
    //     console.log(`Caught an exception: ${err}`);
    //     console.log('Problem with waitForAllDispatches.');
    // } finally {
    //     console.log("2nd try at waitForAllDispatches");
    //     await handler.waitForAllDispatches();
    // }
    console.log('Done waiting');
    await handler.close();
}

// my data gets passed in once an agent is available
async function getDatasetCost(agent, dataset) {
    try {
    var { name, href } = dataset;
        // console.log(`here1: ${href}`);
        if (!href.startsWith('http')) href = `https://ulixee.org${href}`;
        console.log(href);
        await agent.goto(href);
        await agent.waitForPaintingStable( { timeoutMs: 5000 } );
        // await agent.activeTab.waitForPaintingStable( { timeoutMs: 5000 } );
        // await agent.activeTab.waitForLoad('PaintingStable', { timeoutMs: 5000 } );
        // await agent.activeTab.waitForLoad('DomContentLoaded');
        const cost = await agent.document.querySelector('.cost .large-text').textContent;
        console.log('Cost of %s is %s', dataset.name, cost);
    } catch (err) {
        console.log(`Caught an exception: ${err}`);
        console.log(`Problem with getDatasetCost: ${name}, ${href}`);
        // process.exit();
    }
  }

run();
```

and server code:

```
// process.env.DEBUG = true;

// Set this ENV variable to make sure secret-agent knows this is a headless server, so don't try to start Replay
process.env.SA_SHOW_REPLAY="false";
process.env.NODE_DISABLE_COLORS = 1;

const Core = require('@secret-agent/core');

(async () => {
  Core.onShutdown = () => {
    console.log('Exiting Core Process');
    process.exit();
  };
  console.log('Starting Core Process');
  try {
    await Core.start({ coreServerPort: 2600 });
  } catch (err) {
    console.log(`Caught an exception: ${err}`);
    console.log('Problem with Core.start');
    process.exit(1);
  }
})();
```

blakebyrnes:
This PR adds the following features to handlers to manage connections coming and going.

- Agent.coreHost - retrieve the coreHost that an agent is connected to.
- Handler.addConnectionToCore(options | connectionObject) - connect to a new handler
- Handler.removeConnectionToCore(coreHost: string) - remove and disconnect connection to core
- Handler will throw DisconnectedFromCoreError if an agent shuts down/disconnects from a CoreConnection while still mid-operation.

closes #165, #163

### DIFF
--- a/client/connections/ConnectionToCore.ts
+++ b/client/connections/ConnectionToCore.ts
@@ -14,6 +14,7 @@ import CoreSession from '../lib/CoreSession';
 import { IAgentCreateOptions } from '../index';
 import Agent from '../lib/Agent';
 import CoreSessions from '../lib/CoreSessions';
+import DisconnectedFromCoreError from './DisconnectedFromCoreError';
 
 const { log } = Log(module);
 
@@ -49,6 +50,7 @@ export default abstract class ConnectionToCore {
     } else {
       this.hostOrError = Promise.resolve(new Error('No host provided'));
     }
+    this.disconnect = this.disconnect.bind(this);
   }
 
   protected abstract internalSendRequest(payload: ICoreRequestPayload): Promise<void>;
@@ -81,18 +83,25 @@ export default abstract class ConnectionToCore {
   public async disconnect(fatalError?: Error): Promise<void> {
     if (this.isClosing) return;
     this.isClosing = true;
-    const logid = log.stats('ConnectionToCore.Disconnecting');
+    const logid = log.stats('ConnectionToCore.Disconnecting', {
+      host: this.hostOrError,
+      sessionId: null,
+    });
 
-    this.cancelPendingRequests();
+    await this.cancelPendingRequests();
     if (this.connectPromise) {
-      await this.internalSendRequestAndWait({
-        command: 'disconnect',
-        args: [fatalError],
-      });
+      await this.internalSendRequestAndWait(
+        {
+          command: 'disconnect',
+          args: [fatalError],
+        },
+        2e3,
+      );
     }
     await this.destroyConnection();
-    log.stats('RemoteConnectionToCore.Disconnected', {
+    log.stats('ConnectionToCore.Disconnected', {
       parentLogId: logid,
+      host: this.hostOrError,
       sessionId: null,
     });
   }
@@ -155,21 +164,32 @@ export default abstract class ConnectionToCore {
 
   protected async internalSendRequestAndWait(
     payload: Omit<ICoreRequestPayload, 'messageId'>,
+    timeoutMs?: number,
   ): Promise<ICoreResponsePayload> {
-    const { promise, id } = this.createPendingResult();
+    const { promise, id, resolve } = this.createPendingResult();
+
+    let timeout: NodeJS.Timeout;
+    if (timeoutMs) timeout = setTimeout(() => resolve(null), timeoutMs).unref();
     try {
       await this.internalSendRequest({
         messageId: id,
         ...payload,
       });
     } catch (error) {
+      clearTimeout(timeout);
       if (error instanceof CanceledPromiseError) {
         this.pendingRequestsById.delete(id);
         return;
       }
       throw error;
     }
-    return promise;
+
+    // now run to completion with timeout
+    try {
+      return await promise;
+    } finally {
+      clearTimeout(timeout);
+    }
   }
 
   protected onEvent(payload: ICoreEventPayload): void {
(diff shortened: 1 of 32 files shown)


client/connections/ConnectionToCore.ts
```diff
@@ -81,18 +83,25 @@ export default abstract class ConnectionToCore {
...
+        {
+          command: 'disconnect',
+          args: [fatalError],
+        },
+        2e3,
+      );
     }
     await this.destroyConnection();
-    log.stats('RemoteConnectionToCore.Disconnected', {
+    log.stats('ConnectionToCore.Disconnected', {
```
calebjclark, line 102: Why drop "Remote"? Just curious.

blakebyrnes: This one was mis-labeled. It's inside ConnectionToCore, not RemoteConnectionToCore.

PR #177, opened 2021-02-24
Timeout bugs

blakebyrnes:
- Mitm could throw unhandled errors during write if the stream got closed
- Tab DomContentLoaded can take a really long time if you loop and re-load in the same tab a lot of times. Extended the timeout to 60 seconds

### DIFF
--- a/commons/SourceMapSupport.ts
+++ b/commons/SourceMapSupport.ts
@@ -1,6 +1,6 @@
-import * as SourceMapSupport from 'source-map-support';
+import { install } from 'source-map-support';
 
 if (!Error[Symbol.for('source-map-support')]) {
   Error[Symbol.for('source-map-support')] = true;
-  SourceMapSupport.install({ uncaughtShimInstalled: false });
+  install({ handleUncaughtExceptions: false });
 }
--- a/core/lib/TabNavigationsObserver.ts
+++ b/core/lib/TabNavigationsObserver.ts
@@ -118,7 +118,7 @@ export default class TabNavigationsObserver {
   public cancelWaiting(cancelMessage: string): void {
     clearTimeout(this.waitingForLoadTimeout);
     for (const promise of [this.waitingForResource, this.waitingForPromise]) {
-      if (!promise || !promise.isResolved) continue;
+      if (!promise || promise.isResolved) continue;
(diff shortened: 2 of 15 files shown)


commons/SourceMapSupport.ts
```diff
@@ -1,6 +1,6 @@
-import * as SourceMapSupport from 'source-map-support';
+import { install } from 'source-map-support';
 
 if (!Error[Symbol.for('source-map-support')]) {
   Error[Symbol.for('source-map-support')] = true;
-  SourceMapSupport.install({ uncaughtShimInstalled: false });
+  install({ handleUncaughtExceptions: false });
```
calebjclark, line 5: I find that destructured method/property imports to be much harder to follow in the code base. I realize that this allows for smaller files, but does it matter for a server-side tool like SA?

core/lib/TabNavigationsObserver.ts
```diff
@@ -221,7 +221,7 @@ export default class TabNavigationsObserver {
     if (this.waitingForPromise) this.cancelWaiting('New location trigger created');
 
     this.waitingForStatus = status;
-    this.waitingForPromise = createPromise<void>(timeoutMs ?? 30e3);
+    this.waitingForPromise = createPromise<void>(timeoutMs ?? 60e3);
```
calebjclark, line 224: double the default timeout?

commons/SourceMapSupport.ts
blakebyrnes: No, it doesn't really matter, but in this specific case, source map support was behaving strangely with the es export * and nothing else is happening in this file, so I thought an exception was ok.

core/lib/TabNavigationsObserver.ts
blakebyrnes: DOMContentLoaded has been recorded taking 34 and 32 seconds when I was hitting the same url over and over.... so I moved to 60. I don't like having a timeout here with a specific value, but it kind of needs to be something.

commons/SourceMapSupport.ts
calebjclark: Cool. That's a good reason.

PR #182, opened 2021-03-09
Modify Browser<=>Mitm Waiting, Fix Shared Workers, Fix Replay, OOM Fix

blakebyrnes:
This PR changes the following:
1. Previously, we were waiting for http requests to show up in Chrome before allowing them to continue in the mitm. This causes issues for things like Shared Workers, which Chrome won't see in all cases. The new pattern is to "match" chrome resources to the Mitm after they've resolved. I believe this approach will fix some "stalling" issues that we see occasionally
2. SharedWorker emulation works correctly now
3. I disabled the Mitm cache for now. I think we need to re-think this a bit.

Bugs:
1. Out of memory error was coming from Bezierjs. I pulled in the small portion of this library that we needed for ghost
2. Replay was allowing Body tags to be removed
3. Cookies were not properly matching domains


core/models/DevtoolsMessagesTable.ts
```diff
@@ -78,9 +78,9 @@ export default class DevtoolsMessagesTable extends SqliteTable<IDevtoolsMessageR
         return `${value.substr(0, 250)}... truncated ${params.payload.length - 250} chars`;
       }
 
-      // clean out post data and headers (we have these in resources table)
+      // clean out post data (we have these in resources table)
       if ((key === 'headers' || key === 'postData') && params.request) {
-        return undefined;
+        return 'SA_REMOVED_FOR_DB';
```
calebjclark, line 83: What is this string? Should we have a const list somewhere?

blakebyrnes: It's just used to put in the db so you know why it was removed. Didn't seem to deserve a const besides the minor memory saving.

PR #183, opened 2021-03-10
feat(replay): update for mac silicon + deps

blakebyrnes:
Also update deps for dependabot, and synchronize some other deps


replay/backend/managers/OverlayManager.ts
```diff
@@ -37,11 +37,11 @@ export default class OverlayManager {
   }
 
   public destroy = () => {
-    this.browserViews.forEach(x => x.destroy());
+    this.browserViews.length = 0;
   };
 
   public sendToAll = (channel: string, ...args: any[]) => {
-    this.browserViews.forEach(x => !x.isDestroyed() && x.webContents.send(channel, ...args));
+    this.browserViews.forEach(x => x?.webContents.send(channel, ...args));
```
calebjclark, line 44: What prompted this change?

calebjclark: Is it due to the electron upgrade?

blakebyrnes: Electron got rid of these commands. They "say" it operates like a normal js object now.

PR #186, opened 2021-03-16
fix(puppet): remove disable-gpu flag

blakebyrnes:
The current flags we have turned on for GPU are disabling gpu and gpu compositing on Macs, Windows and Linux. I think we originally turned the flags on this way to that a docker/machine with no gpu would run the same way as a development machine. The problem is some sites (like https://github.com) with a lot of webgl are incredibly slow when you do that.

I'm proposing (not 100% sure on this) that we should NOT disable gpu by default. NOTE that disable gpu was a workaround for old chromium issues that made gpu not work in headless on win/mac/linux even when there was hardware present (https://crbug.com/737678 and https://crbug.com/729961). Those are resolved at this point, so it's really just about working with no gpu. 

There's also a potential issue with Linux headless using acceleration (https://github.com/chromium/chromium/commit/19671359ae25aa1e30bde90f8ff92453eeaac2ba). I am torn on if we should use this setting on linux, or just allow this to be added on by users if needed. You can add flags by running
```Chrome83.engine.extraLaunchArgs.push('--use-gl=egl').```

NOTE that Puppeteer and Playwright have both removed disable-gpu and use-gl flags at this point.

@calebjclark any thoughts? Should we add a condition to add use-gl=egl to linux? My only concern is it appears to disable using a software solution and force a hardware one, so it potentially creates new issues on machines without gpus.

### DIFF
--- a/core/lib/GlobalPool.ts
+++ b/core/lib/GlobalPool.ts
@@ -5,7 +5,7 @@ import { createPromise } from '@secret-agent/commons/utils';
 import Log from '@secret-agent/commons/Logger';
 import { MitmProxy as MitmServer } from '@secret-agent/mitm';
 import ICreateSessionOptions from '@secret-agent/core-interfaces/ICreateSessionOptions';
-import Puppet from '@secret-agent/puppet';
+import Puppet, { ILaunchArgs } from '@secret-agent/puppet';
 import * as Os from 'os';
 import IBrowserEngine from '@secret-agent/core-interfaces/IBrowserEngine';
 import DefaultBrowser from '@secret-agent/emulate-chrome-83';
@@ -28,6 +28,7 @@ export default class GlobalPool {
     return this.activeSessionCount < GlobalPool.maxConcurrentAgentsCount;
   }
 
+  private static defaultLaunchArgs: ILaunchArgs;
   private static _activeSessionCount = 0;
   private static puppets: Puppet[] = [];
   private static mitmServer: MitmServer;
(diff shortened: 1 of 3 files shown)


A-Posthuman:
I do primarily run secret-agent on gpu-less linux cloud instances, so I hope you continue to allow some type of config to let that work.

blakebyrnes:
Looks like the windows tests failed, but linux was fine in the github actions. It looks like chrome was unable to open new tabs on Windows.  I'm guessing there's no gpu on these test boxes.. 

I need to check "which" version of chrome should have fixed the need for disable-gpu on windows.

blakebyrnes:
I have a TODO: to test out the gpu/disable-gpu flags on a headless/gpu-less amazon box.

One possible way to approach this is to add bootup parameters to core - something like disableGpu as a Core.start param.

Another is we could have more generic chrome args that can be passed into Core.start.

blakebyrnes:
It does NOT appear to be straightforward to checking if a gpu exists in nodejs, but I think we could likely have no disable-gpu by default, and "if" the first page on a browser fails to launch, we could check if Chrome logs an error about not being able to locate the gpu module. If we see that, we could re-launch Chrome with disable-gpu. This, combined with the ability to pass in a parameter to disable-gpu might be the right combo for this.

A-Posthuman:
That sounds like it could work. I'm just starting to play with using this in Docker, so in that case I'd just like to have a way to pass in a parameter when it starts.

blakebyrnes:
I've tested now on all 3 platforms headless, mac headed, and docker. These configurations seem to work well, but in case they don't, we have an SA_DISABLE_GPU flag that can now be enabled to turn off ALL rasterization and gpu work (ie, don't even run the gpu thread).

PR #193, opened 2021-03-19
fix(emulator): polyfill setAppBadge/clearAppBadge

blakebyrnes:
Calls to clearAppBadge and setAppBadge were crashing Puppet/Chrome. It's a known issue in headless, that appears to be resolved in chrome 89 (https://chromium.googlesource.com/chromium/src.git/+/c8a9fc64753d988e7ee59045a52aea14773fbe0d)

As part of this, I'm proposing we log Chrome stdout and stderr to the database by default, not with additional logging. It was hard to find our crash issue without it.

### DIFF
--- a/core/lib/GlobalPool.ts
+++ b/core/lib/GlobalPool.ts
@@ -107,11 +107,9 @@ export default class GlobalPool {
     this.puppets.push(puppet);
 
     const showBrowser = Boolean(JSON.parse(process.env.SHOW_BROWSER ?? 'false'));
-    const showBrowserLogs = Boolean(JSON.parse(process.env.DEBUG ?? 'false'));
     const browserOrError = await puppet.start({
       proxyPort: this.mitmServer.port,
       showBrowser,
-      pipeBrowserIo: showBrowserLogs,
     });
     if (browserOrError instanceof Error) throw browserOrError;
     return puppet;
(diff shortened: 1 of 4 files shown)


calebjclark:
Let's discus this PR...

The changes in `emulate-browsers/base/injected-scripts/navigator.ts`, which are only needed in Chrome 88 and below, is a good reason why I believe using `@secret-agent/emulate-browsers-base` as a common dependency for all emulators is an icky approach.

blakebyrnes:
Sure, we can discuss. I don't think this changes my previous position though. This is mixed into another file because I was trying not to conflict with your ongoing (29 days old!!) work on getting the rest of the emulators generated. I figured we could come back and put app badge into it's own file, and only reference this one up to chrome 89. That said, it doesn't have any negative effect (beyond extra work) to put it into chrome 89. 

Given I don't have access to a "slab" project where I can change Chrome 89 to stop adding a particular script, I went in the direction of just adding it into the navigator script.

FWIW, to summarize my memory of our previous discussions about this:
1. You've proposed we move the maintenance of these scripts into Slab and auto-copy them into each monorepo project
2. Your thought was this will reduce conflicting versions, where updating a polyfill/script would impact plugins you don't want it to
  - List time we talked this through, we came to the conclusion that point 2 is not actually a problem, because you a plugin refers to a specific emulator-base version, so the scripts will not change out underneath it.
3. I'm concerned with the maintenance of copying all these files into tons of emulators, and/or checking in tons of auto-generated code. 
4. I also dislike the thought of managing these in a different repo. Maybe I can get used to it, or we can have one folder for "active" scripts that's different than the ones copied into each emulator
5. I think we can work around these issues by simply creating different "versions" of the scripts, and then importing the appropriate ones when we generate the slab plugins. I find this less confusing to think about than the idea that different emulators actually reference different "versions" of the git history of each script in slab.

calebjclark:
Ha. Ok. End of discussion.

PR #207, opened 2021-03-31
Mitm bug, install deps message issue

blakebyrnes:
2 bug fixes
1. The message to run the 'install dependencies' script broke from showing the original message in core. Writing a unit test for this led to me to an unhandled rejection issue.
2. When the mitm re-uses http1 connections, they can disconnect before they get reused, but the socket doesn't know. This fix tests out the socket before reusing it (by attempting to send the http request and seeing if an econnreset happens)

### DIFF
--- a/client/connections/ConnectionToCore.ts
+++ b/client/connections/ConnectionToCore.ts
@@ -181,7 +181,7 @@ export default abstract class ConnectionToCore extends TypedEventEmitter<{
     return this.isDisconnecting === false && this.coreSessions.hasAvailability();
   }
 
-  public async createSession(options: ICreateSessionOptions): Promise<CoreSession | Error> {
+  public async createSession(options: ICreateSessionOptions): Promise<CoreSession> {
     try {
       const sessionMeta = await this.commandQueue.run<ISessionMeta>('Session.create', options);
       const session = new CoreSession({ ...sessionMeta, sessionName: options.sessionName }, this);
@@ -189,7 +189,7 @@ export default abstract class ConnectionToCore extends TypedEventEmitter<{
       return session;
     } catch (error) {
       if (error instanceof DisconnectedFromCoreError && this.isDisconnecting) return null;
-      return error;
+      throw error;
     }
   }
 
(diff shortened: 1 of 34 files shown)


client/connections/ConnectionToCore.ts
```diff
@@ -181,15 +181,15 @@ export default abstract class ConnectionToCore extends TypedEventEmitter<{
...
+  public async createSession(options: ICreateSessionOptions): Promise<CoreSession | Error> {
     try {
       const sessionMeta = await this.commandQueue.run<ISessionMeta>('Session.create', options);
       const session = new CoreSession({ ...sessionMeta, sessionName: options.sessionName }, this);
       this.coreSessions.track(session);
       return session;
     } catch (error) {
       if (error instanceof DisconnectedFromCoreError && this.isDisconnecting) return null;
-      throw error;
+      return error;
```
calebjclark, line 192: Is this a new pattern? Returning errors instead of throwing them?

blakebyrnes: It's not new... this is part of the unfortunate reality that if you throw this error and no one is looking, node shits itself.

calebjclark: But you're catching the errors down in the Agent class.

blakebyrnes: I was trying to track down an unhandled exception. I can switch this back and re-catch in Agent if you want to minimize locations we do this weird thing.

commons/test/TypeSerializer.test.ts
```diff
@@ -2,9 +2,9 @@ import Puppet from '@secret-agent/puppet';
 import Chrome83 from '@secret-agent/emulate-chrome-83';
 import TypeSerializer, { stringifiedTypeSerializerClass } from '../TypeSerializer';
 import { CanceledPromiseError } from '../interfaces/IPendingWaitEvent';
-import Log from '../Logger';
+import logger from '../Logger';
```
calebjclark, line 5: Doesn't this go against our convention of naming the import variable the same name as the file?

blakebyrnes: Yeah.. I exported the class "Log" from this class, and suddenly eslint started complaining about 3 files where we did things like this. All our logger imports go against several conventions. We use this code everywhere, but the default export is actually a logger function, which should be named createLogger. I think at some point we just need a quick cleanup of the logger exports and imports.

`import Log from '@secret-agent/commons/Logger';`

PR #215, opened 2021-04-13
fix(puppet): fix chrome-88

blakebyrnes:
Fixes Chrome 88. It seems like Chrome 88 pauses the debugger on new tabs, and won't allow it through until you run Runtime.runDebuggerIfWaiting. This means our current strategy of giving emulators access to puppet to perform initialization will hang. I think we need to adjust so that a page simply provides "scripts" to be run on each tab.

### DIFF
--- a/client/lib/Agent.ts
+++ b/client/lib/Agent.ts
@@ -329,6 +329,13 @@ export default class Agent extends AwaitedEventTarget<{ close: void }> {
       return onrejected(err);
     }
   }
+
+  public toJSON(): any {
+    // return empty so we can
+    return {
+      type: 'Agent',
+    };
+  }
 }
 
 // This class will lazily connect to core on first access of the tab properties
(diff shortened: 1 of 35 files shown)


client/lib/Agent.ts
```diff
@@ -329,6 +329,13 @@ export default class Agent extends AwaitedEventTarget<{ close: void }> {
       return onrejected(err);
     }
   }
+
+  public toJSON(): any {
+    // return empty so we can
+    return {
+      type: 'Agent',
```
calebjclark, line 336: Why not use `this.constructor.name`?

blakebyrnes: There's not a base class. Are you thinking it would just be better if someone copies and pastes a whole class?

calebjclark: It's just a safety precaution in case the class gets renamed. The string her is unlikely to be caught.

blakebyrnes: Ok, fine by me. You want to just update when you pull it in?

blakebyrnes: FWIW, I considered making this more of an instructional message - like, toJSON in SecretAgent doesn't attempt to serialize all properties since they are async

PR #225, opened 2021-04-26
Only chrome latest

calebjclark:
All other emulators have moved to their own repos.

emulate-chrome-88 is a submodule in order for chrome-latest to pull in during development mode.


core/test/user-profile.test.ts
```diff
@@ -133,49 +132,6 @@ describe('UserProfile cookie tests', () => {
       await tab.close();
     }
   });
-
-  it('restores cookies for safari', async () => {
```
blakebyrnes, line 137: Do you think we'll just throw out this test? Or would it move into the safari repo?

yarn.lock
```diff
@@ -2684,6 +2684,31 @@
...
 
+"@secret-agent/core-interfaces@1.4.1-alpha.4":
+  version "1.4.1-alpha.4"
+  resolved "https://registry.yarnpkg.com/@secret-agent/core-interfaces/-/core-interfaces-1.4.1-alpha.4.tgz#912792be401298b747981a75786c86d7622af393"
+  integrity sha512-cge9C5NhLEVU5b8aBniPBGwwTjtqSDOVS+vi+GeNmQlsfcC97s9bMxXymBWi/WU/5rAY7Tn4UfsoJLvQfqIwgg==
+  dependencies:
+    "@secret-agent/puppet-interfaces" "1.4.1-alpha.4"
+    awaited-dom "^1.1.12"
+
+"@secret-agent/emulate-safari-13@1.4.1-alpha.4":
```
blakebyrnes, line 2695: Should this dependency be here?

yarn.lock
```diff
@@ -2684,6 +2684,31 @@
   dependencies:
     "@octokit/openapi-types" "^5.3.2"
 
+"@secret-agent/core-interfaces@1.4.1-alpha.4":
```
blakebyrnes, line 2687: Didn't this get removed?

blakebyrnes (commented):
Do you have a sense of how we'll manage dependencies for the new emulator packages? I assume lerna won't work across the repos? Right now, it looks like the emulators depend on puppet, interfaces and commons (for logger and createPromise). If we introduce a new version of puppet, any sense of how we can maintain compatibility? Will the new repos watch the main branch and test against them or something?

calebjclark:
I haven't fully figured that out.... my current thought is we test emulators against the main SA repo using yarn workspaces from the emulator repos.

yarn.lock
calebjclark: Removed

yarn.lock
calebjclark: Removed

core/test/user-profile.test.ts
calebjclark: I moved it into the safari repo.

PR #245, opened 2021-05-11
fix(dns): handle dns socket disconnect

blakebyrnes:
If an EOF is received from the dns socket, we need to try to reconnect

### DIFF
--- a/mitm/lib/DnsOverTlsSocket.ts
+++ b/mitm/lib/DnsOverTlsSocket.ts
@@ -5,6 +5,7 @@ import IResolvablePromise from '@secret-agent/interfaces/IResolvablePromise';
 import { createPromise } from '@secret-agent/commons/utils';
 import MitmSocket from '@secret-agent/mitm-socket/index';
 import { CanceledPromiseError } from '@secret-agent/commons/interfaces/IPendingWaitEvent';
+import { addTypedEventListener, removeEventListeners } from '@secret-agent/commons/eventUtils';
 import RequestSession from '../handlers/RequestSession';
 
 export default class DnsOverTlsSocket {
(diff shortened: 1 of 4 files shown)


mitm/lib/DnsOverTlsSocket.ts
```diff
@@ -5,6 +5,8 @@ import IResolvablePromise from '@secret-agent/interfaces/IResolvablePromise';
 import { createPromise } from '@secret-agent/commons/utils';
 import MitmSocket from '@secret-agent/mitm-socket/index';
 import { CanceledPromiseError } from '@secret-agent/commons/interfaces/IPendingWaitEvent';
+import { addTypedEventListener, removeEventListeners } from '@secret-agent/commons/eventUtils';
+import Resolvable from '@secret-agent/commons/Resolvable';
```
calebjclark, line 9: Unused

mitm/test/dns.test.ts
```diff
@@ -83,6 +83,18 @@ describe('DnsOverTlsSocket', () => {
       socket.close();
     }
   });
+
+  // this test has to wait for the upstream to disconnect, so skip by default
+  test.skip('should be able to lookup a record after a miss', async () => {
```
calebjclark, line 88: Why have this in here if we're skipping it?

blakebyrnes: Good question. That's how I tested that it was disconnecting, and then worked if the original socket disconnected. I guess I just didn't want to throw it away if i needed to test it again?

calebjclark: Hrm... yeah, I'm not sure.

PR #246, opened 2021-05-12
Move Replay resources + dom replay scripts into core; Disable Mitm

blakebyrnes:
This PR does some maintenance on Replay and Core.

1. Move replay scripts to Core API and serve them as resources. This allows Core to version scripts to the data, as well as to prepare us for being able to replay dom inside Core/scraping scripts
2. Allow the MITM to be disabled and capture resources accordingly
3. Compress the DOMChanges so they can be streamed more quickly
4. Add back AllContentLoaded internally - we had some tests that need to wait for load


core/injected-scripts/domReplayer.ts
```diff
@@ -0,0 +1,410 @@
...
+    replayInteractions(...args: any[]);
+    getIsMainFrame?: () => boolean;
+    idMap: Map<number, Node>;
+    debugLogs: any[];
+    debugToConsole: boolean;
+    getNodeById(id: number): Node;
+  }
+}
+
+// copied since we can't import data types
```
calebjclark, line 17: Why can't data types be imported?

blakebyrnes: This is an injected script - gets run in the browser...

core/lib/FrameNavigationsObserver.ts
```diff
@@ -162,14 +166,11 @@ export default class FrameNavigationsObserver {
     for (const state of this.navigations.top.stateChanges.keys()) {
       // don't resolve states for redirected
       if (state === LocationStatus.HttpRedirected) continue;
-      let recordedStatus = PipelineStatus[state as IPipelineStatus];
-
-      // use painting stable "order of pipeline" if the passed in state is load
+      let pipelineStatus = PipelineStatus[state as IPipelineStatus];
       if (state === LoadStatus.Load) {
-        recordedStatus = PipelineStatus.PaintingStable;
+        pipelineStatus = PipelineStatus.AllContentLoaded;
```
calebjclark, line 171: Why the switch from PaintingStable to AllContentLoaded?

core/injected-scripts/domReplayer.ts
calebjclark: So it's converted to javascript, right? And typescript's enums are converted to a JS object.

core/lib/FrameNavigationsObserver.ts
blakebyrnes: Painting Stable actually gets handled separately right above this snippet. This is assigning a pipeline status to the "load" event that gets stored in the load statuses.

PR #257, opened 2021-06-05
WIP: fix proxy leaks

blakebyrnes:
This PR isn't intended to merge, but to share the approach. I think we should put this into the new plugin architecture. Any thoughts on how far to go with the "Debugger" idea though?

NOTE: this hasn't been hardened or unit tested yet. I need to make sure it works in flattened, minified scripts where breakpoints aren't really by "line".

### DIFF
--- a/core/index.ts
+++ b/core/index.ts
@@ -24,8 +24,8 @@ export default class Core {
   public static server = new CoreServer();
   public static readonly connections: ConnectionToClient[] = [];
   public static pluginMap: {
-    humanEmulatorsById: { [id: string]: IHumanEmulatorClass }
-    browserEmulatorsById: { [id: string]: IBrowserEmulatorClass }
+    humanEmulatorsById: { [id: string]: IHumanEmulatorClass };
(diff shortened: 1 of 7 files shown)


blakebyrnes:
@calebjclark this is rewritten to live in a plugin... however... it slows down the creepjs page from about 54ms to run the suite of property checks to 28 seconds... I'm unsure what I did. It was running faster previously, but now seems to hang up chrome.

calebjclark:
Is the 54ms to 28 seconds a comparable test on the same version? i.e., are you sure it's the switch to becoming a plugin that is slowing this branch down? There have been a substantial number of code changes over the past few weeks.

blakebyrnes:
54ms to 28seconds was just from turning on/off the debugger actually breaking. It seems to send Chrome into a cpu spin (not much logs during that period). I was getting around 10 seconds return before we did the last round of changes, but I don't know what's causing it. Could be something new in the plugins (we have some new random timeouts in tests), could be that creepjs changed their code, or it could be one of the other changes

PR #267, opened 2021-06-29
feat: dynamically require core extender plugins required by client

calebjclark:
ClientExtender plugins are allowed to specify an array of CoreExtender plugin IDs that it requires (see execute-js-plugin). However, those same CoreExtender plugins must be installed in Core. This can be frustrating/difficult since by default the Core is started automatically and in another process -- therefore, the developer has no easy access to adding the CoreExtender through Core.use().

This PR allows Core to automatically require the CoreExtender so long as it available through the node require. As a security precaution it checks to ensure the package is actually a CoreExtender.

Also, Core now includes a allowDynamicPluginDependencies static property that can be set to false (but is true by default). Setting this property to false disables the ability to dynamically require these CoreExtender dependencies.

### DIFF
--- a/core/lib/Plugins.ts
+++ b/core/lib/Plugins.ts
@@ -20,6 +20,7 @@ import IPlugins from '@secret-agent/interfaces/IPlugins';
 import IPlugin, { IPluginClass } from '@secret-agent/interfaces/IPlugin';
 import IPluginCreateOptions from '@secret-agent/interfaces/IPluginCreateOptions';
 import IBrowserEngine from '@secret-agent/interfaces/IBrowserEngine';
+import { PluginTypes } from '@secret-agent/interfaces/IPluginTypes';
 import Core from '../index';
 
 const DefaultBrowserEmulatorId = 'default-browser-emulator';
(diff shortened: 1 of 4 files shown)


core/lib/Plugins.ts
```diff
@@ -164,11 +190,26 @@ export default class Plugins implements IPlugins {
...
   private addPluginInstance(plugin: IPlugin) {
     this.instances.push(plugin);
     this.instanceById[plugin.id] = plugin;
   }
+
+  private require(pluginId: string): IPluginClass {
+    if (!Plugins.pluginClassesById[pluginId]) {
+      try {
+        // eslint-disable-next-line global-require,import/no-dynamic-require
+        const Plugin = require(pluginId);
```
blakebyrnes, line 206: Does this need to have the .default check to handle typescript generated default exports?

blakebyrnes:
I was kind of wondered if we'd end up packaging plugins with databoxes. This makes sense though... pretty annoying to get it to load in another process

core/lib/Plugins.ts
calebjclark: Yes. Will push a new commit.

calebjclark: Pushed.

PR #268, opened 2021-06-29
fix(plugins): inconsistent dom env properties

blakebyrnes:
This PR fixes some creepjs issues, and some structural changes caused by the plugins that make deviceMemory and enumerated devices rotate per dom environment.

### DIFF
--- a/plugins/default-browser-emulator/index.ts
+++ b/plugins/default-browser-emulator/index.ts
@@ -14,6 +14,7 @@ import IPluginCreateOptions from '@secret-agent/interfaces/IPluginCreateOptions'
 import IUserAgentOption from '@secret-agent/interfaces/IUserAgentOption';
 import BrowserEngine from '@secret-agent/plugin-utils/lib/BrowserEngine';
 import IGeolocation from '@secret-agent/interfaces/IGeolocation';
+import { randomBytes } from 'crypto';
 import Viewports from './lib/Viewports';
 import setWorkerDomOverrides from './lib/setWorkerDomOverrides';
 import setPageDomOverrides from './lib/setPageDomOverrides';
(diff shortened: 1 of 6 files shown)


plugins/default-browser-emulator/index.ts
```diff
@@ -51,7 +52,19 @@ export default class DefaultBrowserEmulator extends BrowserEmulatorBase {
...
+    this.data.videoDevice = {
+      deviceId: randomBytes(32).toString('hex'),
+      groupId: randomBytes(32).toString('hex'),
+    };
+    this.data.webGLParameters = {
+      // UNMASKED_VENDOR_WEBGL
+      37445: 'Intel Inc.',
+      // UNMASKED_RENDERER_WEBGL
+      37446: 'Intel Iris OpenGL Engine',
+    };
```
calebjclark, line 67: This new code seems to fundamentally changes the clarity of `data` from being the "imported emulator data files for a browser" to something that we can hang extra data on if we don't know a better place to put it.

I'm not against the change, and I definitely don't have a better alternative at the moment, so I'm inclined to go with it. Just putting it out there in case it spurs some fresh ideas.

blakebyrnes: I kind of recognized that you were doing that and kind of didnt’. Maybe if we called it ‘generatedData’ or something, that would help at least differentiate?
2 options I can think of:
1. We can make the emulator builder back into a single instance per plugin instance. You changed it to be created everytime a new page or new worker is created. I wasn’t entirely sure why, but I remember you mentioning it, so I left it
2. We could have the idea of a “userProfile” or “computerProfile” include things like this - device id, gpu, memory, etc. we could have it be the second parameter into those helpers

PR #269, opened 2021-06-29
feat(core): improve mouse message when not there

blakebyrnes:
Improves the messages when a dom node can't be hovered prior to clicking.


plugins/default-human-emulator/index.ts
```diff
@@ -140,14 +140,27 @@ export default class DefaultHumanEmulator extends HumanEmulatorBase {
       ? await helper.lookupBoundingRect([targetRect.nodeId], true, true)
```
calebjclark, line 140: Line 14 above (IMouseUpResult) is defined but never used.

PR #277, opened 2021-07-01
fix(puppet): clean data dir on close

blakebyrnes:
When running "headed", if the data directory is not cleaned up, it could result in not being able to launch a second time.


puppet/lib/launchProcess.ts
```diff
@@ -109,6 +115,7 @@ export default async function launchProcess(
       } else {
         launchedProcess.kill('SIGKILL');
       }
+      if (dataDir) Fs.rmdirSync(dataDir, { recursive: true });
```
calebjclark, line 118: I've noticed that rmdirSync throws errors every so often. I can't figure out why. We might want a try-catch block that retries x times.

PR #280, opened 2021-07-04
First stab at documenting plugins + simplified plugin architecture

calebjclark:
The more I tried to document the new plugin architecture (and provide examples), the more it became apparent that the plugin structure was too complex.

This is my first stab at reducing some of the complexity.

- Client Extenders are now just Client Plugins
- Core Extenders are now just Core Plugins
- agent.use can now any of the following: 
1) Client Plugin object, 
2) an object that contains (among other things) ClientPlugins -- see how `@secret-agent/execute-js` includes both Client and Core plugins.
3) An npm package string (i.e., `@secret-agent/execute-js`). If a Core Plugin is found in the package then it's passed through to Core and loaded there as well.
4) An absolute path to a file that exports one or more Client or Core plugins. Similar to #3, if a Core plugin is found then it's passed through to Core and loaded there. This is helpful during local development of a plugin or if you want to create a custom one for your project -- just use node's `require.resolve('./relative-path')` which dynamically returns the absolute.
- Similar to agent.use, Core.use receives both objects (Core plugins) or strings (npm packages or absolute paths). This also allows you to use require.resolve in a Client plugin's coreDependencyIds property during local development.
- Client and Core `.use()` previously used different logic flows for loading their respective plugins. Both now operate the exact same using the shared `@secret-agent/plugin-utils`.
- Core contains a property called `allowDynamicPluginLoading`, which is true by default. Setting this to false disables the ablity to pass plugin dependencies from `agent.use` through to `Core.use` for dynamic require. Instead, an error will be thrown if the plugin hasn't been explicitly required by Core.
- The `pluginType` property for Client and Core Plugins has been renamed to just `type`. It was unnecessary clutter since it was already attached to the plugin object (`plugin.type`).
- I flipped the relationship between Browser/Human Emulators and Core Plugins. Previously all three were distinctly different in their own respective ways. In the simplified approach, Browser/Human Emulators just extend Core Plugins. You should just trust me on this -- it was too confusing before. The old approach wasn't really explainable in any understandable way, which I quickly found out when trying to document it. ha.
- A first stab at plugin documentation is ready to go. I skipped on some things, like documenting HumanEmulators since I'm leaning towards depreciating them (the same functionality is in Core Plugins).


website/docs/BasicInterfaces/Handler.md
```diff
@@ -146,7 +145,7 @@ NOTE: when using this method, you must call [`agent.close()`](/docs/basic-interf
 
 - options `object`. Accepts any of the following:
   - name `string`. This is used to generate a unique sessionName.
-  - browserEmulatorId `string` defaults to `chrome-latest`. Emulates a specific browser engine version.
+  - browserEmulatorId `string` defaults to `default-browser-emulator`. Emulates the properties that help SecretAgent look like a normal browser.
   - humanEmulatorId `string`. Drives human-like mouse/keyboard movements.
```
blakebyrnes, line 149: Should we remove mention of this while it's in limbo?

website/docs/BasicInterfaces/Agent.md
```diff
@@ -74,7 +74,7 @@ const { Agent } = require('secret-agent');
   - connectionToCore `options | ConnectionToCore`. An object containing `IConnectionToCoreOptions` used to connect, or an already created `ConnectionToCore` instance. Defaults to automatically booting up and connecting to a local `Core`.
   - name `string`. This is used to generate a unique sessionName.
   - userAgent `strong`. This sets your browser's user agent string. Prefixing this string with a tilde (~) allows for dynamic options.
-  - browserEmulatorId `string`. Emulates the unique properties that help SecretAgent look like a normal browser.
+  - browserEmulatorId `string`. Emulates the properties that help SecretAgent look like a normal browser.
```
blakebyrnes, line 77: There's a more detailed description under "handler" that seems to give a bit more info on what to actually do with this

website/docs/Plugins/CorePlugins.md
```diff
@@ -0,0 +1,161 @@
...
+  - viewport `Viewport`. This is an object containing browser width and height as well as screenWidth and screenHeight, among other properties.
+  - geolocation `Geolocation`. This is an object containing longtitude and latitude, among other properties.  
+  - timezoneId `string`. The configured unicode TimezoneId or host default (eg, America/New_York).
+  - locale `string`. The configured locale in use (eg, en-US).
+
+Modify any value in the object to change it session-wide.
+
+#### **Returns** `void`
+
+### onClientCommand<em>(meta, ...args)</em> *optional*
```
blakebyrnes, line 89: Could use a client-side example here. It's hard to wrap your head around how this works/what you're supposed to do.

website/docs/BasicInterfaces/Handler.md
calebjclark: remove mention of humanEmulatorId?

website/docs/Plugins/CorePlugins.md
calebjclark: The example at the top of the page is a onClientCommand example.

website/docs/BasicInterfaces/Handler.md
blakebyrnes: Wondering if we should. It seems unclear what to do with this parameter in the current setup.

blakebyrnes: Like, is it now a humanEmulator"PluginId"?

calebjclark:
How is it any different than before? Or are you suggesting we go ahead and
remove HumanEmulator as a concept?

website/docs/BasicInterfaces/Handler.md
calebjclark: How is it any different than before? Or are you suggesting we go ahead and
remove HumanEmulator as a concept?

blakebyrnes: I'm not sure I have a good thought on fully removing. I do think there are times you'd want to direct a given scrape to use a specific human profile. It seems like a difference from before is we had a page explaining the 3 built-in human emulators you could switch between, and you installed additional ones into "HumanEmulators.use". Now, you have to install a plugin, and then refer to the pluginId? So you're not really referring to a humanEmulatorId, but a pluginId, right?

calebjclark: A humanEmulatorId is a pluginId as in a HumanEmulator extends CorePlugin.

blakebyrnes: So maybe the docs should at least be updated to say "The pluginId that should be used to drive human-like mouse/keyboards movements for the session". And did you say there was a way to disable the default plugin? Like passing null or something?

calebjclark: Fixed.

website/docs/BasicInterfaces/Agent.md
calebjclark: Fixed

PR #287, opened 2021-07-14
WIP: renamed SA to Hero

calebjclark:
This is fully working and tested, except for two things:
- It's expecting to find asset releases for MitM and Replay to be a github.com/ulixee/ulixee, which doesn't exist right now.
- It's expecting to find chrome-versions on the @ulixee npm namespace, which also doesn't exist.


Dockerfile
```diff
@@ -38,27 +38,27 @@ ENV PATH $GOPATH/bin:/usr/local/go/bin:$PATH
...
-    && chown -R sagent:sagent /app/secret-agent \
-    && mv ~/.cache/secret-agent /home/sagent/.cache/ \
+    && chown -R sagent:sagent /app/ulixee \
+    && mv ~/.cache/ulixee /home/sagent/.cache/ \
     && chmod 777 /tmp \
-    && chmod -R 777 /home/sagent/.cache/secret-agent
+    && chmod -R 777 /home/sagent/.cache/ulixee
 
 # Add below to run as unprivileged user.
 USER sagent
```
blakebyrnes, line 64: should prolly rename this too, right?

hero/full-client/index.cjs
```diff
@@ -1,3 +0,0 @@
-require('./index.js');
```
blakebyrnes, line 1: Why delete the commonjs support? Are we gonna get rid of default exports?

hero/core/lib/InjectedScripts.ts
```diff
@@ -32,7 +32,7 @@ const injectedScript = `(function installInjectedScripts() {
        ${pageScripts.pageEventsRecorder}
     })('${pageEventsCallbackName}');
 
-    window.SA = {
+    window.ULX = {
```
blakebyrnes, line 35: Seems like these should be Hero, not ULX

hero/examples/core-server.mjs
```diff
@@ -1,11 +1,11 @@
-import Core from '@secret-agent/core';
+import Core from '@ulixee/hero-core';
 
 (async () => {
   Core.onShutdown = () => {
     console.log('Exiting Core Process');
     process.exit();
   };
-  await Core.start({ coreServerPort: 1337 }, !process.env.SA_TEMPORARY_CORE);
+  await Core.start({ coreServerPort: 1337 }, !process.env.ULX_TEMPORARY_CORE);
```
blakebyrnes, line 8: I know there's a TON of stuff here, but I'm finding the times we're using ULX vs Hero to be ambiguous

Dockerfile
calebjclark: done

hero/full-client/index.cjs
calebjclark: It's not deleted. That might have been related to the rename from full-client to main. I reverted it back to full-client.

hero/examples/core-server.mjs
calebjclark: I believe it's just for env vars, which I agree can just as well be HERO_.

hero/full-client/index.cjs
blakebyrnes: Oh, I meant to delete this comment after I saw main

hero/examples/core-server.mjs
calebjclark: Updated

PR #288, opened 2021-07-14
Restructure Fixes + Catalina for Chrome 88/89

blakebyrnes:


hero/examples/news.ycombinator.com.ts
```diff
@@ -3,7 +3,7 @@ import agent, { Observable } from 'secret-agent';
 // process.env.SA_SHOW_BROWSER = 'true';
 
 async function run() {
-  await agent.configure({ userAgent: '~ chrome >= 88' });
+  await agent.configure({ userAgent: '~ chrome = 88.*' });
```
calebjclark, line 6: Why the asterisk?

lerna.json
```diff
@@ -5,6 +5,7 @@
     "hero/client",
     "hero/core",
     "hero/interfaces",
+    "here/plugin-utils",
```
calebjclark, line 8: I believe this should be "hero" not "here".

hero/examples/news.ycombinator.com.ts
blakebyrnes: compareVersions otherwise says nothing = 88.0.0

lerna.json
blakebyrnes: Doh

hero/examples/news.ycombinator.com.ts
blakebyrnes: Of Chrome? No... The versions are like 88.0.4324.{many patches}

blakebyrnes: It works without the asterisk, so I removed it from the example.

PR #298, opened 2021-07-16
fix: missing hero variable renames

blakebyrnes:


core/injected-scripts/interactReplayer.ts
```diff
@@ -278,29 +277,29 @@ function updateScroll(scrollEvent: IFrontendScrollRecord) {
 /////// BUILD UI ELEMENTS //////////////////////////////////////////////////////////////////////////////////////////////
 
 let isInitialized = false;
-function createReplayItems() {
+function createReplayItems(showChromeAlive = false) {
```
calebjclark, line 280: This seems to be the only Chrome Alive code.

PR #299, opened 2021-07-17
Simplified Hero

calebjclark:
This PR primarily does three things:

-  No default instance (`hero`) is exported from the `@ulixee/hero`. Instead it operates how all the other tools will operate -- it exports the class:
```
import Hero from '@ulixee/hero';

const hero = new Hero();
```

- Handler has been moved from Hero to Runner. It's expected that most developers will use Runner for high-throughput queries, but `p-queue` or other libraries can be used to build their own tooling.  I updated examples/ulixee.org.ts to use `p-queue` instead of Handler. 

- `@ulixee/hero-full-client` has been replaced with `@ulixee/hero-fullstack`. The new fullstack loads Core in the same process and then uses LocalConnectionToCore to directly interface with it. By default values between Client and Core are not serialized, but LocalConnectionToCore.shouldSerialize allows a developer to toggle this.

- Core's http server has been moved to Server. If someone wants to run Hero without Server then the recommended approach is with `@ulixee/hero-fullstack`. Or they can write their own http interface to Core.


client/connections/ConnectionFactory.ts
```diff
@@ -1,33 +1,34 @@
 import Log from '@ulixee/commons/Logger';
 import IConnectionToCoreOptions from '../interfaces/IConnectionToCoreOptions';
 import ConnectionToCore from './ConnectionToCore';
-import RemoteConnectionToCore from './RemoteConnectionToCore';
+import RemoteServerConnectionToCore from './RemoteServerConnectionToCore';
```
blakebyrnes, line 4: What do you think of calling this ConnectionToRemoteCoreServer? I REALLY liked our switch to ConnectionToClient and ConnectionToReplay in terms of having the name of what you connect to at the end.

fullstack/lib/LocalConnectionToCore.ts
```diff
@@ -0,0 +1,35 @@
+import Core from '@ulixee/hero-core';
+import { ConnectionToCore } from '@ulixee/hero';
+import ICoreRequestPayload from '@ulixee/hero-interfaces/ICoreRequestPayload';
+import ConnectionToClient from '@ulixee/hero-core/connections/ConnectionToClient';
+import TypeSerializer from '@ulixee/commons/TypeSerializer';
+
+export default class LocalConnectionToCore extends ConnectionToCore {
```
blakebyrnes, line 7: If we change to what I proposed above, this would probably be ConnectionToLocalCore

client/connections/ConnectionFactory.ts
calebjclark: Done

fullstack/lib/LocalConnectionToCore.ts
calebjclark: Agreed. Pushed your change.

PR #301, opened 2021-07-18
Moved output from Hero to Databox

calebjclark:
+ I renamed @ulixee/testing to @ulixee/hero-testing as I've come to the conclusion that each tool needs its own testing package.


client/lib/Hero.ts
```diff
@@ -465,8 +452,7 @@ class SessionConnection {
   constructor(private hero: Hero, stateOptions: IStateOptions) {
     const { connectionToCore, ...options } = stateOptions;
 
-    // @ts-ignore
-    const createConnectionToCoreFn: ICreateConnectionToCoreFn = hero.constructor.createConnectionToCore;
+    const createConnectionToCoreFn = (hero.constructor as any).createConnectionToCore;
```
blakebyrnes, line 455: I don't understand this code. Is it just saying `Hero.createConnectionToCore`? You can't change it per instance right?

calebjclark: Look at fullstack/lib/Hero.ts. It extends the Hero class and sets ConnectionToLocalCore on createConnectionToCore. Therefore, I can't refer to Hero.createConnectionToCore. I must respect the child class.

fullstack/index.ts
```diff
@@ -1,6 +1,5 @@
 import '@ulixee/commons/SourceMapSupport';
 import {
-  Observable,
   ConnectionToRemoteCoreServer,
   ConnectionToCore,
   InteractionCommand,
```
blakebyrnes, line 5: Lint error a few lines down. On purpose? We should add the eslint hint if so

client/lib/Hero.ts
blakebyrnes: Oh, I see. Should we have a definition on the base class that just doesn't have an implementation? Kind of seems like an abstract variable of sorts.

client/lib/Hero.ts
```diff
@@ -410,6 +409,45 @@ export default class Hero extends AwaitedEventTarget<{ close: void }> implements
...
+    listenerFn: (event: IClassEvents[K]) => any,
+  ): void {
+    return this.removeListener(eventType, listenerFn);
+  }
+
+  public static once<K extends keyof IClassEvents>(
+    eventType: K,
+    listenerFn: (event: IClassEvents[K]) => any,
+  ): void {
+    return this.addListener(eventType, listenerFn);
```
blakebyrnes, line 449: Should this actually implement once?

client/lib/Hero.ts
calebjclark: I added to IHero.

calebjclark: but yes, I will add.

calebjclark: Good idea, will add.

client/lib/Hero.ts
calebjclark: Databox won't implement once, but why shouldn't the functionality be there?

client/lib/Hero.ts
calebjclark: Added

fullstack/index.ts
calebjclark: Fixed, but it's not showing in my Github UI.

client/lib/Hero.ts
blakebyrnes: It just needs to call this.emitter.once... is my point ;)

calebjclark: Ah. Good catch. Pushed fix.

PR #302, opened 2021-07-21
Hero Fixes

blakebyrnes:
Various fixes related to using headed and hero branch
- fullstack not exporting correct variables
- convert new Date().getTime() to Date.now where appropriate
- Properly close track closed puppets
- Don't hang shutting down chrome pages when browser crashes
- Properly disconnect from http2 sessions that "goaway"
- Save preferences for devtools (like preserve log, dock mode, etc)
- Don't log socket "data" packets sent unless specifically requested

Added GlobalPool events to listen for all browsers closed and all windows closed in all browsers.
- all-browsers-closed: this is needed so we know when we should shut down fullstack
- browser-closed-all-windows: this is needed in headed mode and fullstack so we can tell if the user has killed off all browser windows and just wants to close things down

### DIFF
--- a/client/lib/Hero.ts
+++ b/client/lib/Hero.ts
@@ -50,7 +50,7 @@ import ScriptInstance from './ScriptInstance';
 import AwaitedEventTarget from './AwaitedEventTarget';
 import IHeroDefaults from '../interfaces/IHeroDefaults';
 import CoreSession from './CoreSession';
-import ConnectionFactory, { ICreateConnectionToCoreFn } from "../connections/ConnectionFactory";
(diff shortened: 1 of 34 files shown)


client/lib/Hero.ts
```diff
@@ -497,8 +498,23 @@ class SessionConnection {
       createConnectionToCoreFn,
     );
 
+    if (connection !== connectionToCore) {
+      this.didCreateConnection = true;
```
calebjclark, line 502: Isn't this more of a `didConnect`?

blakebyrnes: I don't think so. It's not related to if connect actually gets called. More, did this hero create it's own "ConnectionToCore", which we need to clean up after

PR #306, opened 2021-07-23
feat(core): keepalive and resume a session

blakebyrnes:
This feature enables us to keep alive a Chrome session and resume it.

In headed, we include user interactions and a bar showing your current command (so you can see the changes happening).

SessionResume works by comparing the sequence of commands you provide to a previous run. If we can pick up where you left the browser, we ONLY play the new commands. This helps you to build out a new script without constantly running from the beginning.

Here are the different ways SessionResume works (arguments):
- sessionId: 
  - if kept alive, it will resume the current browser context based on the start location
  - if closed, will load the session attributes from the db and recreate it with the same settings
- startLocation: where to resume commands
  - currentLocation: tries to run with as many commands that match as possible. Once a command "forks" the rest of the commands will run (vs returning the previous result)
  - pageStart: reload all commands beginning at the currently loaded page (until commands don't match)
  - sessionStart: reset cookies/storage state in the current browser window and start the script over. Will also load from the db if the session is not active anymore.

### DIFF
--- a/core/lib/FrameEnvironment.ts
+++ b/core/lib/FrameEnvironment.ts
@@ -241,6 +241,7 @@ export default class FrameEnvironment {
       result: entry.result,
       runStartDate,
       endDate,
+      run: this.session.resumeCounter,
     };
     if (this.sessionState.nextCommandMeta) {
       const { commandId, sendDate, startDate: clientStartDate } = this.sessionState.nextCommandMeta;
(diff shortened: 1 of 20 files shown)


core/lib/FrameEnvironment.ts
```diff
@@ -241,6 +241,7 @@ export default class FrameEnvironment {
       result: entry.result,
       runStartDate,
       endDate,
+      run: this.session.resumeCounter,
```
calebjclark, line 244: I find this `run` key to be very vague and confusing.

blakebyrnes: Is it the name "run" or just that it doesn't have "id" on it or "counter"? Like, would "runId" make it clearer?

